PR #76, opened 2015-01-21
Closed-Captions feature

LukaszGrela:
This adds the ability to use closed captions and set multiple language version, the _media has the cc object with list of languages, the `HBS` template is using the `#each` helper to enumerate and create track tags, key's of cc element are used as `srclang` value.

### DIFF
--- a/templates/media.hbs
+++ b/templates/media.hbs
@@ -15,8 +15,8 @@
             <source src="{{_media.mp4}}" type="video/mp4"/>
             <source src="{{_media.ogv}}" type="video/ogg"/>
           {{/if}}
-          {{#if _media.subtitles}}
-            {{#each _media.subtitles}} 
+          {{#if _media.cc}}
+            {{#each _media.cc}} 
             <track kind="subtitles" src="{{this}}" srclang="{{@key}}" />
             {{/each}}
           {{/if}}          
--- a/templates/media.hbs
+++ b/templates/media.hbs
@@ -15,11 +15,11 @@
             <source src="{{_media.mp4}}" type="video/mp4"/>
             <source src="{{_media.ogv}}" type="video/ogg"/>
           {{/if}}
-           {{#if _useClosedCaptions}}
-            {{#each _media.cc}} 
-            <track kind="subtitles" src="{{this}}" srclang="{{@key}}" />
-            {{/each}}
-          {{/if}}          
+          {{#if _useClosedCaptions}}
+              {{#each _media.cc}}
+                  <track kind="subtitles" src="{{src}}" srclang="{{srclang}}" />
+              {{/each}}
+          {{/if}}
         </video>
       {{/if}}
     {{/if}}
--- a/example.json
+++ b/example.json
@@ -10,6 +10,7 @@
 	"body": "This is optional body text. Select the play button to start the video.",
 	"instruction": "",
 	"_setCompletionOn":"inview",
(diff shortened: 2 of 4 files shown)


oliverfoster:
can you rename the variable 'lFeatures'  to just 'features'?

+1

templates/media.hbs
```diff
@@ -15,8 +15,13 @@
             <source src="{{_media.mp4}}" type="video/mp4"/>
             <source src="{{_media.ogv}}" type="video/ogg"/>
           {{/if}}
+          {{#if _media.cc}}
```
hrajotia: Would be good if you use `{{#if _useClosedCaptions}}` here.

LukaszGrela: true that:) it was before introduction of this flag:)

hrajotia: for below case -
```
"_useClosedCaptions": true,
"_media": {
	"cc": {
	}
}
```

this `{{#each _media.cc}}` code won't produce any error as `#each` helper won't iterator over empty object `_media.cc`.

Although the better approach would be something like -
```
"_useClosedCaptions": true,
"_media": {
	"cc": [
		{
			"srclang": "en",
			"src": "course/en/video/01_01_040_sub.srt"
		}
	]
}
```
and 
```
{{#if _useClosedCaptions}}
	{{#each _media.cc}}
		<track kind="subtitles" src="{{src}}" srclang="{{srclang}}" />
	{{/each}}
{{/if}}
```
what do you think?

LukaszGrela: I was thinking for the content developer to reduce his typing:) but I think some of the language codes may not be best for the object keys so your approach will be better. - I'll adjust.

hrajotia: The array of objects approach seems to me more convenient because the key part explain us the role of data it holds which makes the model more understable.

Also within few weeks or month we're going to release the authoring tool then there will be less typing for content developers too ;)

LukaszGrela: OK I'll modify it:)

taylortom:
+1

moloko:
If this is going to go directly into master, should we get a proper QA done first? What's the standard in this situation?

moloko:
given that the updates to contrib-spoor caused an error in the authoring tool, a test to make sure any updates to master don't cause issues with the authoring tool might seem sensible.

taylortom:
Yep, sounds good to me. I can get something booked in?

moloko:
that would be great, thanks Tom

Might be a good idea to also give QA a copy of a course with the pre-update media player in it, so they can ensure that they are not reporting issues that already exist ;-)

zenduo:
Would it not make sense for this to be worked into the develop branch rather than the master?

zenduo:
I just dropped this into a fresh "master" course and when building received the following error in my build... causing it to fail.


originalError:
   { [Error: Parse error using esprima for file: /Users/Alan/Desktop/GitHub/my-adapt-course/src/components/adapt-contrib-media/js/adapt-contrib-media.js
   Error: Line 43: Unexpected token ;]

LukaszGrela:
Thanks for spotting, I still have this habit to close line with semicolon, which in JS is not always (like here) valid

taylortom:
I agree with @zenduo that this isn't really suitable for master

zenduo:
@LukaszGrela would you mind removing this pull request and submitting a new one to the develop branch?

PR #83, opened 2015-02-10
Force Block level completion

LukaszGrela:
Feature to set block level completion - useful when there is few media components and just anyone should trigger completion.

This should be used with the "_setCompletionOn":"play", as the function is called in the `onCompletion` handler only.

### DIFF
--- a/example.json
+++ b/example.json
@@ -10,6 +10,7 @@
 	"body": "This is optional body text. Select the play button to start the video.",
 	"instruction": "",
 	"_setCompletionOn":"inview",
+        "_forceBlockCompletion":true,
 	"_media": {
 		"mp4": "course/assets/big_buck_bunny.mp4",
 		"ogv": "course/assets/big_buck_bunny.ogv",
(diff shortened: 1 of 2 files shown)


taylortom:
If this is a feature that people want, then let's add it. I've never heard of any course we've done have this requirement though. Can we float this on the forum or in a separate request ticket?

LukaszGrela:
:) yes, it is from the course I've done recently

moloko:
I can see the sense in this, it just feels like this isn't the best place for it. I feel more like it should actually be set at block level, via a config setting like:
```
_requireCompletionOf: ['all','any','c-05','c-10']
```
If you see what I mean?
So, in @LukaszGrela's case it would be set to:
```
_requireCompletionOf: ['any']
```

LukaszGrela:
I can see your point @moloko, I did what was best for the course. When I will have some time I will try to add it to the block functionality.

taylortom:
I like @moloko's suggestion. Shall we close this PR?

LukaszGrela:
This is now implemented with this PR https://github.com/adaptlearning/adapt_framework/pull/545

PR #84, opened 2015-03-12
Accessibility merge

oliverfoster:


taylortom:
+1

dancgray:
+1, does .media-top-padding ever get used?

oliverfoster:
Not automatically no. It has to be specified. Much like adding a < br > to the _body attribute (as it was in the component.json). Instead you'd need to add media-top-padding to the _classes attribute. It's only for when you need that extra padding to bring the media component inline with an adjacent text component.

taylortom:
+1

moloko:
+1

Issue #72, opened 2014-11-26
make the additional settings of MediaElement available via JSON


moloko:
probably best to have a generic 'options' object so that if we change to a different underlying player, or stick with MediaElement and it changes an option name or adds an option, we don't have to update our code.

LukaszGrela:
The easiest would be to expose features list, component code would validate entries and translate into correct versions if changed (old version props).

    "_features":[],

then in code


    validateFeatures: function(features) {
        if(features)
        {
            var valid = [];
            //process validation
            //...
            return valid;
        }
        else
        {
            //use default
            return ['playpause','progress','current','duration'];
        }
    }
    //...
    postRender: function(){
        //...
        features: this.validateFeatures(this.model.get("_features"))
        //...
    }

oliverfoster:
+1

PR #85, opened 2015-03-24
Issue/#72

taylortom:
Exposes player options to JSON data, fixes https://github.com/adaptlearning/adapt-contrib-media/issues/72


LukaszGrela:
Can this be extended so there is a global media player setup and then per component? Imagine 3 or more media components with the same setup, and you have to put the same code 3 times where you could put "defaults" in the `config.json` and only override per component

    {
        "_defaultLanguage": "en",
        "_questionWeight": 1,
        "_drawer": {
            "_showEasing": "easeOutQuart",
            "_hideEasing": "easeInQuart",
            "_duration": 400
        },
        "_spoor": {
        <CUT FOR BREVITY>
        },
        "_playerOptions": {
            "defaultVideoWidth": 480,
            "defaultVideoHeight": 270,
            "videoWidth": -1,
            "videoHeight": -1,
            "audioWidth": 400,
            "audioHeight": 30,
            "startVolume": 0.8,
            "loop": false,
            "enableAutosize": true,
            "features": ["playpause","progress","current","duration","tracks","volume","fullscreen"],
            "alwaysShowControls": false,
            "iPadUseNativeControls": false,
            "iPhoneUseNativeControls": false,
            "AndroidUseNativeControls": false,
            "alwaysShowHours": false,
            "showTimecodeFrameCount": false,
            "framesPerSecond": 25,
            "enableKeyboard": true,
            "pauseOtherPlayers": true,
            "keyActions": []
        }
    }

LukaszGrela:
Also in relation to closed caption features this option would be useful too:
`toggleCaptionsButtonWhenOnlyOne:true,`

moloko:
@LukaszGrela the global config idea is a really interesting one - not just for this component but perhaps could apply to all components. I think that needs thinking about properly though, particularly with respect to how it would appear in the authoring tool. @taylortom is that something you could raise with the authoring tool development team? 

as for ```toggleCaptionsButtonWhenOnlyOne``` - I don't see that as an available option of MediaElement, which is really what this PR is all about - making sure that the 'player options' listed on the [MediaElement](http://mediaelementjs.com/) page are all configurable from the .json

taylortom:
I really like the global defaults idea, but agree that config.json maybe isn't the best place for it. Will see what the other guys think

LukaszGrela:
@taylortom `config.json` was an example as `spoor` plugin already is using it as a settings holder.
@moloko what do you mean by "I don't see [`toggleCaptionsButtonWhenOnlyOne`] as an available option of MediaElement"? It is available, see here: https://github.com/johndyer/mediaelement/blob/2.13.1/src/js/mep-feature-tracks.js#L14 I think this was just not updated on the web page.

moloko:
@LukaszGrela I stand corrected! @taylortom can you add this in?

taylortom:
Sure, does the following make sense:
``` javascript
"_media": {
   "_playerDefaults": {
      ...
   }
}
```

moloko:
sorry I meant add in ```toggleCaptionsButtonWhenOnlyOne```

taylortom:
I'm not really sure what I need to do here, is it just update the README/example.json?
I've set the code up to pass the whole settings object when initialising the MEP, so as long as the setting's supported and in the json, it should work (see https://github.com/adaptlearning/adapt-contrib-media/pull/85/files#diff-009c3188529ea82310d31db5d82d0c8fR28)

moloko:
Oh yeah, good point - I suppose there is nothing to do really! Sorry, not with it today

taylortom:
Cool. Does this solution work for you @LukaszGrela?

moloko:
Sorry @taylortom but would this work if ```_playerOptions``` was not in the JSON? I don't think you should have to include it given that 90% of the time you probably would be quite happy with the defaults.

I would also add something to the README explaining about the player options, particularly pointing out that they have all be listed in example.json for completeness, but that you don't necessarily need to include the whole lot, just the property/ies you want to change from the default.

taylortom:
Thos should've been fixed in the last two commits

oliverfoster:
+1

moloko:
certainly were
+1

hrajotia:
+1

PR #86, opened 2015-03-24
Lukasz Grela - Closed Captions feature

LukaszGrela:
Closed caption feature with only the template `hbs` changes so it can fit with the PR https://github.com/adaptlearning/adapt-contrib-media/pull/85 which externalises the features of the media player.

### DIFF
--- a/example.json
+++ b/example.json
@@ -10,13 +10,26 @@
 	"body": "This is optional body text. Select the play button to start the video.",
 	"instruction": "",
 	"_setCompletionOn":"inview",
+	"_useClosedCaptions":true,	
 	"_media": {
 		"mp4": "course/assets/big_buck_bunny.mp4",
 		"ogv": "course/assets/big_buck_bunny.ogv",
 		"poster": "course/assets/big-buck-bunny.jpg",
+		"poster": "course/en/video/big-buck-bunny.jpg",
+            	"cc": [
+            		{
+		         	"srclang": "en",
+		        	"src": "course/en/video/big_buck_bunny_sub.vtt"
+            		},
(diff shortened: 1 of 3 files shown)


zenduo:
+1 (it worked, and one small tweak below)

Hey, just checking this out, 2 things:

Should example.json have 2 poster images attributes?
https://github.com/adaptlearning/adapt-contrib-media/pull/86/files#diff-964a59d491880661889b0e908c47a1e0L16

Also how is the PL version rendered is this detected by the the browser? I noticed that this wasn't an option to change languages?

LukaszGrela:
No automation can be here, this only provides a list of available closed captions, user needs to choose correct from the list.

zenduo:
When you say user, you you mean the end user? e.g. the learner? If so how does the learner select? I couldn't see the option to choose another language?

LukaszGrela:
That's why I said that this change works with the PR#85 when you have ability to change the "features" of the media player, otherwise you need to add tracks into the features list. My previous PR was modifying MEP to do it automatically but it would collide with PR#85.
So once you will set the "tracks" as features you will have the "CC" button in the video controls, if you click then a popup with a list of languages will be displayed (even one language). If you will pass the `toggleCaptionsButtonWhenOnlyOne` into the MEP settings then the CC button will act as toggle switch when only one language defined.

taylortom:
+1

moloko:
+1
I haven't had time to actually test all this myself but @zenduo has and it will all get tested as part of the v2 QA anyway

oliverfoster:
+1

PR #90, opened 2015-04-08
Feature/abu 347

taylortom:
https://adaptlearning.atlassian.net/browse/ABU-347

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -32,9 +32,9 @@ define(function(require) {
 
             var modelOptions = this.model.get('_playerOptions');
 
-            if(!modelOptions.pluginPath) modelOptions.pluginPath = 'assets/';
-            if(!modelOptions.features) modelOptions.features = ['playpause','progress','current','duration'];
-            if(!modelOptions.clickToPlayPause) modelOptions.clickToPlayPause = true;
+            if(modelOptions.pluginPath === undefined) modelOptions.pluginPath = 'assets/';
+            if(modelOptions.features === undefined) modelOptions.features = ['playpause','progress','current','duration'];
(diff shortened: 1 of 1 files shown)


moloko:
+1

oliverfoster:
+1

js/adapt-contrib-media.js
```diff
@@ -54,7 +57,22 @@ define(function(require) {
             }
         },
 
-        // Used to check if the media should reset on revisit
+        // Overrides the default play/pause functionality to stop accidental playing on touch devices
+        setupPlayPauseToggle: function() {
+            this.mediaElement.player.options.clickToPlayPause = false;
```
dancgray: Should we check first to avoid a TypeError before setting this value?

taylortom:
@dancgray would you mind testing https://github.com/taylortom/adapt-contrib-media/tree/feature/ABU-347-error-check?

I can't replicate the issue you describe, so a bit of a stab in the dark.

dancgray:
@taylortom I'm still getting the error. It is occurring on iOS, as before. Do you want me to raise an issue once this (feature/ABU-347 branch) has been merged as it looks like it could be specific to my setup?

taylortom:
Hmm ok. I'll have a test on iOS myself and see if I can figure it out, cheers.

taylortom:
Seems the mediaElement.player is missing on iOS for some reason.

The branch I linked to above fixes this for me, although it means that the code won't execute on iOS, hence making this PR completely useless. 

I'll have a look into other solutions.

dancgray:
+1 seems to fix the issue

oliverfoster:
+1

moloko:
+1

PR #96, opened 2015-05-29
Accessibility fixes

oliverfoster:
Forced buttons to show if accessibility it turned on. 
Fixed issue with controls nested in a visibility:hidden container.


moloko:
+1

oliverfoster:
not quite right

PR #122, opened 2015-09-24
a to button

chetan-hajare:
Since a is changed to button, width gets reduced. So it is set to 100% in media-inline-transcript-button class


moloko:
+1

hrajotia:
+1

lc-thomasberger:
+1

PR #125, opened 2015-09-30
feature/mep-update

danielghost:


moloko:
I'd like to try and give this a proper test. will try to do so today/tomorrow if possible

moloko:
I don't know how big an issue this really is, but this does update does cause problems if you use YouTube video as your source. The player appears and the wrong size and, once you click on it, starts expanding vertically and doesn't stop!

Whether we need to address this this really depends on whether people feel we should be supporting the use of YouTube as the video source or not.

moloko:
Sample JSON for YouTube video:
```
"_media": {
    "source": "//www.youtube.com/watch?v=a1Y73sPHKxw",
    "type": "video/youtube"
},
```

moloko:
Note that YouTube video doesn't seem to work at all in Firefox - in this update or in the current master.

moloko:
if you put the `width` and `height` properties of the <video> tag back into media.hbs then YouTube video works OK in IE9 - but not Chrome (and still not Firefox)

moloko:
Sorry to say this is completely shot in IE8... when you load the page the Flash video player starts expanding in height and just won't stop! Tried putting the `width` and `height` properties of the `<video>` tag back in but this doesn't stop it. Don't know what to suggest I'm afraid...

moloko:
+1
tested on IE8 (XP VM), IE9 (Win 7), IE11 (Win8.1 VM), Safari 9 (Mac OS 10.11), iOS 9 (iPad Mini and iPhone 4S)
checked left and full-width layouts, complete on play, complete on ended, captions, vimeo-as-source, youtube-as-source and fullscreen mode

annoyingly fullscreen mode is still no better than it was before (on desktop anyway, is fine on iPad) and there's [this issue](https://adaptlearning.atlassian.net/browse/ABU-1080) with a <div> overlaying the vimeo controls - but that was present before this update. Finally, [ABU-1022](https://adaptlearning.atlassian.net/browse/ABU-1022) is still present - which isn't really surprising.

moloko:
needs something about `_startLanguage` adding to README but I will do that once this is merged as README updates don't require code reviews

oliverfoster:
@moloko can you retest the vimeo stuff, quick fix to disable mejs controls and use vimeo ones, but seems to work across the browsers.

js/adapt-contrib-media.js
```diff
@@ -29,23 +31,66 @@ define(function(require) {
...
+            switch (media.type) {
+            case "video/vimeo":
+                modelOptions.alwaysShowControls = false;
+                modelOptions.hideVideoControlsOnLoad = true;
+                modelOptions.features = [];
+                if (froogaloopAdded) return callback();
+                Modernizr.load({
+                    load: "assets/froogaloop.js", 
+                    complete: function() {
+                        console.log("vimeo loaded");
```
moloko: need to remove this

moloko:
awesome, vimeo events now working!

just need to remove the `console.log` on line 86 of adapt-contrib-media.js and resolve the conflicts

moloko:
+1

less/mep.less
```diff
@@ -848,7 +864,7 @@
 	left: 0;
 	width: 100%;
 	height: 100%;
-	background: url(../../assets/background.png);
+	background: url(b../../assets/background.png);
```
hrajotia: does the letter 'b' is suppose to be here?

moloko:
+1

hrajotia:
+1

dancgray:
+1

Issue #128, opened 2015-10-20
[develop] adapt-contrib-media - if features are defined then _useClosedCaptions is not used


LukaszGrela:
The lines https://github.com/adaptlearning/adapt-contrib-media/blob/develop/js/adapt-contrib-media.js#L34-L39 wraps the `_useClosedCaptions` test which then is only tested when features are undefined, this should be separated, also defining all features second time is redundant when you can only add "tracks" to the feature list.

```
    if (modelOptions.features === undefined) {
        modelOptions.features = ['playpause', 'progress', 'current', 'duration'];
    }
    //at this point features will not be undefined
    var ccFeature = "tracks";
    if (this.model.get("_useClosedCaptions") && modelOptions.features.indexOf(ccFeature) == -1) {
        modelOptions.features.push(ccFeature);
    }

```

taylortom:
You're right this could be simplified a bit.

```
if (features undefined)
      if(cc) features = ['playpause', 'progress', 'tracks', 'current', 'duration']
      else features = ['playpause', 'progress', 'current', 'duration']
```

I don't really have a problem with the outer `modelOptions.features === undefined` wrapper, as everything inside is features-related.

I'm not sure your proposed solution would be suitable, as I think the indices of the items are used when rendering (i.e. tracks button is needed before the timers in this case) - should probably do a splice here.

taylortom:
Also (thinking about this a bit more), do we want to force-add the CC button if the features list has been specified via JSON?

LukaszGrela:
The use of `_useClosedCaptions` property will force it so if user will not define the features but use this prop should have the tracks added.

PR #129, opened 2015-10-21
Issue #128

LukaszGrela:
split the features test so the tracks are added not only when features is undefined

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -31,12 +31,15 @@ define(function(require) {
             var modelOptions = this.model.get('_playerOptions');
 
             if (modelOptions.pluginPath === undefined) modelOptions.pluginPath = 'assets/';
-            if(modelOptions.features === undefined) {
-                modelOptions.features = ['playpause','progress','current','duration'];
-                if (this.model.get('_useClosedCaptions')) {
-                    modelOptions.features = ['playpause','progress','tracks','current','duration'];
-                }
+            if (modelOptions.features === undefined) {
+                modelOptions.features = ['playpause', 'progress', 'current', 'duration'];
             }
+            //at this point features will not be undefined
+            var ccFeature = "tracks";
+            if (this.model.get("_useClosedCaptions") && modelOptions.features.indexOf(ccFeature) == -1) {
+                modelOptions.features.push(ccFeature);
+            }
+            
             if(modelOptions.clickToPlayPause === undefined) modelOptions.clickToPlayPause = true;
 
             modelOptions.success = _.bind(this.onPlayerReady, this);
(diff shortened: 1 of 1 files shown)


oliverfoster:
``modelOptions.features.indexOf(ccFeature)``  for ie8 needs to be ``_.indexOf(modelOptions.features, ccFeature);``

oliverfoster:
modelOptions.features.push(ccFeature); will always put the cc button at the end after the duration, i'm not sure we want to hardcode the position of the cc button to be at the end always.

oliverfoster:
could you re-include the check for closed captions if ``features`` is not defined and ``_useClosedCaptions`` is enabled?

that way if ``features`` isn't defined and ``_useClosedCaptions`` is enabled, the cc button will appear in a better place. and if ``features`` was defined but tracks wasn't included and ``_useClosedCaptions`` is enabled it'll append the cc button to the end? features should only be used by experienced people and so i'd like the normal defaults to be as easy as possible.

taylortom:
Commented [here](https://github.com/adaptlearning/adapt-contrib-media/issues/128):
'*do we want to force-add the CC button if the features list has been specified via JSON? (this is what removing the undefined check will do)*'

taylortom:
There shouldn't be a conflict there, but still

oliverfoster:
yea, i can understand both reasons, and i think both solutions are fine. just need to re-add the old check in my brain. then all is good.

oliverfoster:
i think injecting the button if features has been included and cc is used is a good idea.

LukaszGrela:
I dont like duplicated work, maybe instead of push I will just inject it at desired place? So adding for undefined features test plus injecting at desired place:
```
if(modelOptions.features === undefined) {
	modelOptions.features = ['playpause','progress','current','duration'];
}

var ccFeature = "tracks";
if(this.model.get("_useClosedCaptions") && _.indexOf(modelOptions.features,ccFeature) == -1){
    var position = _.indexOf(modelOptions.features,'progress');
    if(position!=-1)
    	modelOptions.features.splice(position+1, 0, ccFeature);
    else
        modelOptions.features.push(ccFeature);
}
```

oliverfoster:
there isn't a javascript solution to this problem, it is a human issue with the stack of adapt+media+mejs and the potential want to change the position of the ``cc`` icon.

if someone is using ``features`` they need to understand mediaelementjs as a library and should need to understand that they must include ``tracks`` if they want to include the ``cc`` icon in the player bar.
if they are not using the ``features`` attribute, they should only need to use ``_useClosedCaptions`` and the player button should come out in a chosen default location, as the primary use-case.

injecting the ``cc`` button in an instance where someone is using the ``features`` attribute and the ``_useClosedCaptions`` attribute is not going to work well as you've got a few thousand combinations with the 7 options you can have in features. if someone is using ``features`` i think it is their responsibility to include ``tracks`` if needed.

LukaszGrela:
My premise was that if user is using the features and _useClosedCaptions but skips tracks then one should accept default behaviour, thus there is no permutations involved as it will be either last or after progress. This code behaves like original in results and for defined features but missing tracks and intention of the user was to use tracks through _useClosedCaptions it adds by default, if user is not happy then he just needs to add it to features. So the placement of 'cc' depends only on the user not willing to put tracks in features list and still wanting to have it.

LukaszGrela:
Too much hassle to change something that works

PR #132, opened 2015-12-03
bugfix/hide-controls: default to hide controls

oliverfoster:
hide controls on default and with mouseover unless otherwise specified or accessibility is on


moloko:
hmmm... although I can see this is setting those two properties correctly there doesn't seem to be much visible impact, no matter what I have them set to...! But then I get the same result with the code that's current in master so I guess... +1?!

moloko:
+1

dancgray:
+1

hrajotia:
+1

PR #137, opened 2016-02-29
issue-983: closed captions button not accessible

oliverfoster:
expanded source code and added lines 2334-2353, turns cc button into "click to open" rather than "focus to open" style button, so that the menu can be tabbed onto without it closing. implemented keyboard selection to change language.

https://github.com/adaptlearning/adapt_framework/issues/983


moloko:
I think this might be a good opportunity to swap from the minified version to the full version - as the grunt build process will presumably minify it for us anyway?

If we use the unminified version it'll be so much easier to review this - not to mention easier to see what you did so that we can re-do same should we update the mediaelement code again!

tomgreenfield:
I agree wholeheartedly with Matt. We probably need two commits here; one to add the unminified JS file and another with Ollie's patch.

moloko:
going for the upgrade to mediaelement as well I see!

oliverfoster:
closing to do this a different way

PR #140, opened 2016-04-06
#14

RobertPeek:
A trigger has been added to the default build core notify view, which media-autoplay now listens for to stop the video.


oliverfoster:
Was there anything from here still needed?

RobertPeek:
No.. apologies but this was a mistake, the PR should not have been across the forks. My bad!

PR #141, opened 2016-05-24
Issue-1076

bell87:
Added defensive programming around incorrectly configured media component to avoid infinite spinners on course page load.

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -89,10 +89,9 @@ define(function(require) {
 
         addThirdPartyFixes: function(modelOptions, callback) {
             var media = this.model.get("_media");
-            if (!media) {
-                callback();
-            } else {
-                switch (media.type) {
+            if (!media) return callback();
+            
+            switch (media.type) {
                 case "video/vimeo":
                     modelOptions.alwaysShowControls = false;
                     modelOptions.hideVideoControlsOnLoad = true;
@@ -108,7 +107,6 @@ define(function(require) {
                     break;
                 default:
                     callback();
-                }
             }
         },
 
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -75,7 +75,6 @@ define(function(require) {
                 try {
                     if (this.model.get('_media').source) {
                         this.$('.media-widget').addClass('external-source');
-                        this.setReadyStatus();
                     }
                 } catch (e) {
                     console.log("ERROR! No _media property found in components.json for component " + this.model.get('_id'));
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -81,7 +81,7 @@ define(function(require) {
 
         addMediaTypeClass: function() {
             var media = this.model.get("_media");
-            if (media.type) {
+            if (media && media.type) {
                 var typeClass = media.type.replace(/\//, "-");
                 this.$(".media-widget").addClass(typeClass);
             }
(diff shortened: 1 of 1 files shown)


js/adapt-contrib-media.js
```diff
@@ -81,29 +81,33 @@ define(function(require) {
...
-                Modernizr.load({
-                    load: "assets/froogaloop.js", 
-                    complete: function() {
-                        froogaloopAdded = true;
-                        callback();
-                    }
-                }); 
-                break;
-            default:
+            if (media) {
```
barry-learningpool: I'd invert this and return early to reduce indentation (only a minor thing)

js/adapt-contrib-media.js
```diff
@@ -81,29 +81,33 @@ define(function(require) {
...
+                    Modernizr.load({
+                        load: "assets/froogaloop.js", 
+                        complete: function() {
+                            froogaloopAdded = true;
+                            callback();
+                        }
+                    }); 
+                    break;
+                default:
+                    callback();
```
barry-learningpool: Should these be 'return callback()'?

barry-learningpool:
+1

js/adapt-contrib-media.js
moloko: no, `callback` is a pointer to `createPlayer` which doesn't return a value. From the looks of it, the use of `return callback()` on line 98 is there mainly to get the code to exit `addThirdPartyFixes` at that point

tomgreenfield:
I would agree with @barry-learningpool's suggestion on returning early to sink that indentation.

Rather than:
```js
if (!media) {
    callback();
} else {
    // switch statement
}
```
I was expecting:
```js
if (!media) return callback();

// switch statement
```

dennis-learningpool:
+1

tomgreenfield:
+1

brian-learningpool:
+1

moloko:
+2 (tested in Firefox)

Whilst this fixes the 'infinite spinner' issue it does still leave the author with very little indication of what went wrong other than this error:
unhelpful_error

So (and this is just a suggestion), it might be better to do something like this:
```javascript
// We're streaming - set ready now, as success won't be called above
try {
    if (this.model.get('_media').source) {
        this.$('.media-widget').addClass('external-source');
    }
} catch (e) {
    console.error("ERROR! No _media property found in components.json for component " + this.model.get('_id'));
} finally {
    this.setReadyStatus();
}
```
Which gives you this (in Firefox dev tools):
better_error

What do you think? Sorry, I know you've already got all the votes you need to get this merged... so feel free to ignore this if you don't want to have to get everyone to vote again!

moloko:
And finally a small request - with PRs like these it would help anyone who wants to test if you could give a concrete example of what 'incorrectly configured' looks like. Just saves a bit of time as otherwise it's necessary to work it out from the code!

bell87:
Thanks @moloko that is helpful. Always better to give user feedback where possible. hopefully the latest change is ok and I can get this merged.

moloko:
+2 (testing in FF, Chrome and IE11)

js/adapt-contrib-media.js
```diff
@@ -72,39 +72,47 @@ define(function(require) {
                 this.$('audio, video').mediaelementplayer(modelOptions);
 
                 // We're streaming - set ready now, as success won't be called above
-                if (this.model.get('_media').source) {
-                    this.$('.media-widget').addClass('external-source');
+                try {
+                    if (this.model.get('_media').source) {
+                        this.$('.media-widget').addClass('external-source');
+                        this.setReadyStatus();
```
moloko: this line is redundant due to the `finally` block (anything in a `finally` block will execute regardless of whether there's an error or not)

moloko:
+2

PR #153, opened 2016-09-08
issue-1223-1: memory leak fixes

oliverfoster:
* vimeo cleanup
* missing function call removed
* event and player cleanups


brian-learningpool:
+1

PR #154, opened 2016-09-09
Removed the protocol from streaming services

brian-learningpool:
YouTube and Vimeo give share links with HTTPS, by removing the protocol from the URL it prevents any problems with the protocol not matching the server on which the Adapt content is hosted.


oliverfoster:
+1

moloko:
I'd have put the comments on line 47 (instead of an empty line) but otherwise +1

oliverfoster:
Hang on, does that work with http and https? What's the question mark about?

oliverfoster:
Gotta check my regex...

oliverfoster:
Yup, all good +1 still

hrajotia:
+1

PR #155, opened 2016-09-14
issue-1213: resize window when video height changes 

sharpb:
[adaptlearning/adapt_framework#1213](https://github.com/adaptlearning/adapt_framework/issues/1213)

This fixes an issue we were having using trickle with the media component.  Would value any feedback or better/simpler ways of doing this.  We've tested in the following browsers:
- IE9
- IE11
- Edge
- Chrome (52)
- Firefox (32)

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -123,6 +123,28 @@ define(function(require) {
             } else {
                 this.$('.component-widget').on('inview', _.bind(this.inview, this));
             }
+
+            try {
+                var observer = new MutationObserver(function(mutations) {
+                    mutations.forEach(function(mutation) {
+                        $(window).resize();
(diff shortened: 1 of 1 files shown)


moloko:
sorry but much of this is not compatible with IE8/9/10 which (sadly!) we are still supporting :cry: :sob:

sharpb:
@moloko thanks for the feedback.  I haven't tested in IE8 - I'll give it a go and report back.  The fall back for older browsers isn't pretty but it works :smile: Can't seem to find a better alternative at present.

sharpb:
@moloko I tested in IE8 and attachEventListener caused issues.  So I've updated to use attachEvent in IE8.  Tested and working in the course we're using it in.

moloko:
see comment on original issue

PR #158, opened 2016-10-12
Add autoplay to video and audio

srimko:
We needed to an autoplay feature for a project this PR aims to provide users with this feature.

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -219,6 +219,15 @@ define(function(require) {
                     this.$('.component-inner').off('inview');
                     this.setCompletionStatus();
                 }
+
+                // initialize to false if autoplay option doesn't exist
+                var autoPlay = this.model.get('autoplay') || false;
+                if (this._isVisibleTop && this._isVisibleBottom && autoPlay === true) {
+                    var player = this.mediaElement.player;
+                    if (!player) return;
+
+                    player.play();
+                }
             }
         },
 


moloko:
This is a nice idea, but I had one concern: should it autoplay _every_ time it comes into view or just once? It seems like doing it every time might be a bit annoying, I don't know. I suppose it depends on your use-case for this.

I do think that `_autoPlayOnInview` would be a better name

Don't forget to update example.json, properties.schema and README.md - the latter particularly needs to make it very clear that 'autoplay' is not supported on iOS (possibly others?? need to check) unless user is running iOS10 and the video is silent.

Version number also needs updating in bower.json and README

moloko:
also, the way you have it set up means that autoplay would only work if you also set `_setCompletionOn` to `"inview"`. If you wanted to have `"ended"` you would not also be able to have autoplay

(the component only binds to the inview event when `_setCompletionOn === "inview"`)

srimko:
Ok thank you.
If I want that medias launch himself automatically even if completion is set to`"ended"` or `"play"`.

Is there a function to know if it is visible ?

moloko:
yes, the `inview` event will do this for you, the problem you've got is that it's currently only being registered when `_setCompletionOn` is set to `"inview"` - so you just need to add some code so that you can also listen for `inview` when autoplay is required. Might be best to set up a separate handler function for that rather than try and add it into the existing one?

moloko:
closing due to inactivity

PR #159, opened 2016-11-23
expose media events

lc-thomasberger:
Rather then having global events for media events, I decided to add player state attributes to each model.
One can then listen to the change event of each media component models.

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -160,9 +160,9 @@ define([
             }
 
             // handle other completion events in the event Listeners 
-            this.mediaElement.addEventListener('play', this.onMediaElementPlay);
-            this.mediaElement.addEventListener('pause', this.onMediaElementPause);
-            this.mediaElement.addEventListener('ended', this.onMediaElementEnded);
+            $(this.mediaElement).on('play', this.onMediaElementPlay);
+            $(this.mediaElement).on('pause', this.onMediaElementPause);
+            $(this.mediaElement).on('ended', this.onMediaElementEnded);
         },
 
         onMediaElementPlay: function(event) {
@@ -286,9 +286,9 @@ define([
             }
 
             if (this.mediaElement) {
-                this.mediaElement.removeEventListener('play', this.onMediaElementPlay);
-                this.mediaElement.removeEventListener('pause', this.onMediaElementPause);
-                this.mediaElement.removeEventListener('ended', this.onMediaElementEnded);
+                $(this.mediaElement).off('play', this.onMediaElementPlay);
+                $(this.mediaElement).off('pause', this.onMediaElementPause);
+                $(this.mediaElement).off('ended', this.onMediaElementEnded);
 
                 this.mediaElement.src = "";
                 $(this.mediaElement.pluginElement).remove();
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -171,8 +171,9 @@ define([
                 '_isMediaEnded': false
             });
             
-            if (this.completionEvent === 'play')
+            if (this.completionEvent === 'play') {
                 this.setCompletionStatus();
+            }
         },
 
         onMediaElementPause: function(event) {
@@ -182,8 +183,9 @@ define([
         onMediaElementEnded: function(event) {
             this.model.set('_isMediaEnded', true);
 
-            if (this.completionEvent === 'ended')
+            if (this.completionEvent === 'ended') {
                 this.setCompletionStatus();
+            }
         },
 
         // Overrides the default play/pause functionality to stop accidental playing on touch devices
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -155,13 +155,11 @@ define([
         setupEventListeners: function() {
             this.completionEvent = (!this.model.get('_setCompletionOn')) ? 'play' : this.model.get('_setCompletionOn');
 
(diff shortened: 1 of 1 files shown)


js/adapt-contrib-media.js
```diff
@@ -256,11 +282,17 @@ define([
                     delete mejs.players[player_id];
                 }
             }
+
+            this.mediaElement.removeEventListener('play', this._onMediaElementPlay);
```
moloko: I don't think this works if you've used _.bind on the function because that returns a copy of the original function. You'd have to either store references to the bound functions or, for preference, use [_.bindAll](https://devdocs.io/underscore/index#bindAll) to fix the scope of them all at the same time without creating copies.

js/adapt-contrib-media.js
```diff
@@ -39,6 +39,12 @@ define([
             this.listenTo(Adapt, 'device:changed', this.onDeviceChanged);
             this.listenTo(Adapt, 'accessibility:toggle', this.onAccessibilityToggle);
 
+            // set initial player state attributes
+            this.model.set({
+                '_isMediaEnded': false,
+                '_isMediaPlaying': false
```
brian-learningpool: You're listening for the `pause` event too, so might it be worthwhile to add an `_isMediaPaused` attribute at the same time?

js/adapt-contrib-media.js
```diff
@@ -154,6 +159,27 @@ define([
...
+            this._onMediaElementEnded = _.bind(this.onMediaElementEnded, this);
+
+            this.mediaElement.addEventListener('play', this._onMediaElementPlay);
+            this.mediaElement.addEventListener('pause', this._onMediaElementPause);
+            this.mediaElement.addEventListener('ended', this._onMediaElementEnded);
+        },
+
+        onMediaElementPlay: function(event) {
+            this.model.set('_isMediaPlaying', true);
+            this.model.set('_isMediaEnded', false);
```
brian-learningpool: You could have passed an object for readability, as lines 43-46 above do:
`this.model.set({'_isMediaPlaying', true, '_isMediaEnded', false});`

js/adapt-contrib-media.js
tomgreenfield: Is this not already being done [here](https://github.com/adaptlearning/adapt-contrib-media/blob/f5ef202d035ee6ca305ba270a56c1e21f66508e7/js/adapt-contrib-media.js#L163-L165)?

js/adapt-contrib-media.js
lc-thomasberger: `_isMediaPaused` would just be the inverted version of `_isMediaPlaying`

js/adapt-contrib-media.js
lc-thomasberger: yes, _onMediaElementPlay is the bound version of onMediaElementPlay

moloko: sorry, yes it is. 

if you switch to using _.bindAll you don't need to store references like that

lc-thomasberger:
There is one known limitation, on IE8 the events are not triggered the "first few times".

js/adapt-contrib-media.js
```diff
@@ -256,11 +282,17 @@ define([
                     delete mejs.players[player_id];
                 }
             }
+
+            this.mediaElement.removeEventListener('play', this.onMediaElementPlay);
```
moloko: just to point out that other references to `this.mediaElement` here are within an `if(this.mediaElement)` block - maybe move these down to be inside the one on L290?

moloko:
> There is one known limitation, on IE8 the events are not triggered the "first few times".

I think this is happening because the IE8 API doesn't seem to allow for multiple event handler registrations...? 

The default setup is to have `"play"` as the completion trigger, and if you do then the `onMediaElementPlay` event doesn't get triggered first time. 

Equally if I set `_setCompletionOn` to `"ended"` then `onMediaElementEnded` doesn't fire first time round.

If I set it to `"inview"` then it works fine.

lc-thomasberger:
@moloko I refactored the code to only have one eventListener for each player event. In the handle I set completion and update player state in the object.
Tested in ie8 and it worked.

js/adapt-contrib-media.js
```diff
@@ -148,12 +155,35 @@ define([
...
+            this.mediaElement.addEventListener('ended', this.onMediaElementEnded);
+        },
+
+        onMediaElementPlay: function(event) {
+            this.model.set({
+                '_isMediaPlaying': true,
+                '_isMediaEnded': false
+            });
+            
+            if (this.completionEvent === 'play')
```
moloko: style guide (and consistency) requires curly braces here...

js/adapt-contrib-media.js
```diff
@@ -148,12 +155,35 @@ define([
...
+        },
+
+        onMediaElementPause: function(event) {
+            this.model.set('_isMediaPlaying', false);
+        },
+
+        onMediaElementEnded: function(event) {
+            this.model.set('_isMediaEnded', true);
+
+            if (this.completionEvent === 'ended')
```
moloko: style guide (and consistency) requires curly braces here...

js/adapt-contrib-media.js
moloko: or all on one line if that's your preference

moloko:
Tested in IE8 with both 'play' and 'ended' completion events, set. All events firing as one would expect.

PR #160, opened 2016-11-23
iOS playsinline 

lc-thomasberger:
add playsinline attribute to allow playing videos inline on iOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ guide the learner’s interaction with the component.
 
 **_allowFullScreen** (boolean): Determines whether fullscreen mode is available or not. Note that changing this setting has no effect in Internet Explorer 9 as this browser does not support fullscreen mode for HTML video.
 
+**_playsinline** (boolean): If set to 'true', videos will play inline on iOS 10 and above.        
+
 **_startLanguage** (string): If using closed captions with multiple languages, use this property to specify which language should be shown by default. The value of this property must match one of the **srclang** values.  
 
 **_media** (object): The media attributes group will contain different values depending on the type of media: video or audio.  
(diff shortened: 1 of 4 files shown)


example.json
```diff
@@ -12,6 +12,7 @@
     "_setCompletionOn":"inview",
     "_useClosedCaptions":true,
     "_startLanguage":"en",
+    "_iosPlaysinline": false,
```
moloko: I would just go for `playsinline` as a name and explain in the README that it really only applies to iOS (10 and later) phones. That's the name that they used in the YouTube Player API so it would make it consistent with that.
Oh, and you need to document this new setting in the README and example.json

README.md
```diff
@@ -50,6 +50,8 @@ guide the learner’s interaction with the component.
 
 **_allowFullScreen** (boolean): Determines whether fullscreen mode is available or not. Note that changing this setting has no effect in Internet Explorer 9 as this browser does not support fullscreen mode for HTML video.
 
+**_playsinline** (boolean): If set to 'true', videos will play inline on iOS 10 and above.
```
moloko: Could you change this to "videos will play 'inline' on iPhones (the same way they do on iPads). Note that this feature is only available in iOS10 and above. The default is `false`."

properties.schema
```diff
@@ -172,6 +172,15 @@
...
       "help": "If set to 'true', the full screen button will appear (excluding IE9)."
     },
+    "_playsinline": {
+      "type":"boolean",
+      "required":false,
+      "default": false,
+      "title": "Play videos inline on iOS",
+      "inputType": {"type": "Boolean", "options": [true, false]},
+      "validators": [],
+      "help": "If set to 'true', videos will play inline on iOS 10 and above."
```
moloko: Could you re-word to make it clear this only applies to iPhones?

canstudios-paulh:
+1

PR #161, opened 2016-12-09
Reworded video to media

simondate:
The word video is used several times as an all encompassing word for what could also be an audio file. Should use the generic term `media` instead.

### DIFF
--- a/properties.schema
+++ b/properties.schema
@@ -7,7 +7,7 @@
     "ariaRegion": {
       "type": "string",
       "required": true,
-      "default": "This is a media component which plays media. Select the play / pause button to watch or listen.",
+      "default": "This is a media player component. Select the play / pause button to watch or listen.",
       "inputType": "Text",
       "validators": [],
       "translatable": true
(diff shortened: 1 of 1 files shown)


properties.schema
```diff
@@ -7,7 +7,7 @@
     "ariaRegion": {
       "type": "string",
       "required": true,
-      "default": "This is a media component which plays a video. Select the play / pause button to watch or listen.",
+      "default": "This is a media component which plays media. Select the play / pause button to watch or listen.",
```
moloko: Could you change to 'This is a component which plays media.'?

properties.schema
```diff
@@ -143,7 +143,7 @@
       "title": "Set completion of Video on",
       "inputType": {"type": "Select", "options":["inview", "play", "ended"]},
       "validators": ["required"],
-      "help": "This tells Adapt when to set this video to complete"
+      "help": "This tells Adapt when to set this media to complete"
```
moloko: could you change to: 'Defines what media event should trigger completion of this component.'?

properties.schema
```diff
@@ -143,7 +143,7 @@
       "title": "Set completion of Video on",
```
moloko: I'm thinking 'Completion trigger' might be a better title here?

properties.schema
```diff
@@ -193,7 +193,7 @@
           "title": "Set Completion",
           "inputType": {"type": "Boolean", "options": [true, false]},
           "validators": [],
-          "help": "This tells Adapt to set this video to complete when the inline transcript is shown"
+          "help": "This tells Adapt to set this media to complete when the inline transcript is shown"
```
moloko: Could you change to 'Set to 'true' to have this component mark as completed when the inline transcript is shown.'

properties.schema
```diff
@@ -193,7 +193,7 @@
           "title": "Set Completion",
```
moloko: I think 'Trigger completion?' would be a better title

simondate: With the question mark too?

PR #164, opened 2017-05-11
schema amends: change boolean selects to checkboxes 

moloko:
see adaptlearning/adapt_authoring#1380

**please leave unmerged until all core plugins have been updated**


oliverfoster:
@moloko conflicted ^

PR #165, opened 2017-05-11
Controlled media scrub

AleitaT:
The option '_preventForwardScrubbing' prevents users from skipping to the end of required media components in order to continue on to the next component. If the option '_preventForwardScrubbing' is set to 'true' this prevents users from scrubbing ahead any further than they have already viewed unless the user has completed the unit.  They can freely scrub backwards, and back up until the furthest viewed point.  While there are still possibilities for navigating around this, it removes the most obvious route for users to attempt to circumvent required learning materials.

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -318,9 +318,9 @@ define([
 
             if (this.mediaElement) {
                 $(this.mediaElement).off({
-                    'play': this.onMediaElementPlay,
-                    'pause': this.onMediaElementPause,
-                    'ended': this.onMediaElementEnded,
+                	'play': this.onMediaElementPlay,
+                	'pause': this.onMediaElementPause,
+                	'ended': this.onMediaElementEnded,
                     'seeking': this.onMediaElementSeeking,
                     'timeupdate': this.onMediaElementTimeUpdate
                 });
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -160,6 +160,14 @@ define([
                 this.$('.component-widget').on('inview', _.bind(this.inview, this));
             }
 
+            // wrapper to check if preventForwardScrubbing is turned on.
+            if ((this.model.get('_preventForwardScrubbing')) && (!this.model.get('_isComplete'))) {
+                $(this.mediaElement).on({
+                    'seeking': this.onMediaElementSeeking.bind(this),
+                    'timeupdate': this.onMediaElementTimeUpdate.bind(this)
+                });
+            }
+            
             // handle other completion events in the event Listeners 
             $(this.mediaElement).on({
             	'play': this.onMediaElementPlay,
@@ -190,6 +198,28 @@ define([
                 this.setCompletionStatus();
             }
         },
+        
+        onMediaElementSeeking: function(event) {
+            var maxViewed = this.model.get("_maxViewed");
+            if(!maxViewed){
+                maxViewed = 0;
+            }
+            if(event.target.currentTime > maxViewed){
+                event.target.currentTime = maxViewed;
+            }
+        },
+
+        onMediaElementTimeUpdate: function(event) {
+            var maxViewed = this.model.get("_maxViewed");
+            if(!maxViewed){
+                maxViewed = 0;
+            }
+            if(event.target.currentTime > maxViewed){
+                this.model.set("_maxViewed", event.target.currentTime);
+            }
+        },
+
+
 
         // Overrides the default play/pause functionality to stop accidental playing on touch devices
         setupPlayPauseToggle: function() {
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -40,7 +40,7 @@ define([
             this.listenTo(Adapt, 'device:changed', this.onDeviceChanged);
             this.listenTo(Adapt, 'accessibility:toggle', this.onAccessibilityToggle);
 
-            _.bindAll(this, 'onMediaElementPlay', 'onMediaElementPause', 'onMediaElementEnded');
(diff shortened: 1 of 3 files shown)


js/adapt-contrib-media.js
```diff
@@ -160,6 +160,14 @@ define([
                 this.$('.component-widget').on('inview', _.bind(this.inview, this));
             }
 
+            // wrapper to check if preventForwardScrubbing is turned on.
+            if ((this.model.get('_preventForwardScrubbing')) && (!this.model.get('_isComplete'))) {
+                $(this.mediaElement).on({
+                    'seeking': this.onMediaElementSeeking.bind(this),
```
oliverfoster: can you add these function names to the ```_.bindAll(this``` on line 43 instead of using ```function.bind```?

js/adapt-contrib-media.js
```diff
@@ -190,6 +198,28 @@ define([
                 this.setCompletionStatus();
             }
         },
+        
+        onMediaElementSeeking: function(event) {
+            var maxViewed = this.model.get("_maxViewed");
+            if(!maxViewed){
```
oliverfoster: add spaces:
```javascript
if (!maxViewed) {
```
not
```javascript
if(!maxViewed){
```

js/adapt-contrib-media.js
```diff
@@ -160,6 +160,14 @@ define([
                 this.$('.component-widget').on('inview', _.bind(this.inview, this));
             }
 
+            // wrapper to check if preventForwardScrubbing is turned on.
+            if ((this.model.get('_preventForwardScrubbing')) && (!this.model.get('_isComplete'))) {
+                $(this.mediaElement).on({
```
oliverfoster: you'll need to unbind these functions at the end of the ```remove``` function with the other ```this.mediaElement``` event listeners

oliverfoster:
i really like this

properties.schema
```diff
@@ -181,6 +181,15 @@
       "validators": [],
       "help": "If set to 'true', videos will play 'inline' on iPhones (the same way they do on iPads). Note that this feature is only available in iOS10 and above."
     },
+  "_preventForwardScrubbing": {
+      "type":"boolean",
+      "required":false,
+      "default": false,
+      "title": "Prevents user from skipping ahead in media.",
+      "inputType": {"type": "Boolean", "options": [true, false]},
```
moloko: Would you mind setting this to `"inputType": "Checkbox"`? We're in [the process of changing these over](https://github.com/adaptlearning/adapt_authoring/issues/1380), I've just done a PR (#164) for this component so it would be great if you could make this match - thanks!

moloko: Could you also change the help text to avoid referencing the type of UI control, I've been going with something like: "If enabled, will..."

moloko:
Oooh this is so sneaky... genius!

However, you may have already spotted this, but it doesn't prevent skipping ahead when you're in the full screen player on iOS (which of course iPhones go into by default). You may be able to work around this by accessing the underlying `<video>` element's events rather than targetting `mediaElement`?

Even if you can't work around this, I don't think that should prevent this from going in... we just need to be clear in the README and the properties.schema about this limitation and any others we find. 

I would also be tempted to change the wording from:
> users will not be able to scrub ahead in media
to 
> tries to prevent users from scrubbing ahead in media

js/adapt-contrib-media.js
AleitaT: Added the new functions to the remove call.

properties.schema
AleitaT: Updated to check box, and changed verbiage to match 'enabled' phrasing as well as 'attempt to' since there are still some ways around this.

AleitaT:
Thank you for your feedback.  I believe I've tackled all change requests. This is a feature that is required for the coursework we do, and I hope it benefits others.  I will look forward to exploring options for full screen iOS users in the future.  For now, I hope it satisfies your requirements.

js/adapt-contrib-media.js
```diff
@@ -292,7 +320,9 @@ define([
                 $(this.mediaElement).off({
                 	'play': this.onMediaElementPlay,
                 	'pause': this.onMediaElementPause,
-                	'ended': this.onMediaElementEnded
+                	'ended': this.onMediaElementEnded,
+                    'seeking': this.onMediaElementSeeking,
+                    'timeupdate': this.onMediaElementTimeUpdate
```
oliverfoster: Got a slightly funky bit of whitespace here. Should be spaces 4.

AleitaT: Ok.  There are no tabs remaining in this file now.

properties.schema
```diff
@@ -191,9 +200,9 @@
           "required":false,
           "default": true,
           "title": "Trigger completion?",
-          "inputType": {"type": "Boolean", "options": [true, false]},
+          "inputType": "type": "Checkbox",
           "validators": [],
-          "help": "Set to 'true' to have this component mark as completed when the inline transcript is shown."
+          "help": "If enabled will attempt to prevent users from skipping ahead in audio/video."
```
moloko: Looks like you changed this help text instead of the one on L191 in aa2f926

AleitaT: Sorry about that,  I repaired the _transcript definitions to what they were previously.

properties.schema
```diff
@@ -193,7 +202,7 @@
           "title": "Trigger completion?",
           "inputType": {"type": "Boolean", "options": [true, false]},
           "validators": [],
-          "help": "Set to 'true' to have this component mark as completed when the inline transcript is shown."
+          "help": Set to 'true' to have this component mark as completed when the inline transcript is shown."
```
brian-learningpool: There is a missing quote here.

AleitaT: Well at this point, that's just embarrassing.  Fixed.

PR #167, opened 2017-06-08
Issue/1410

barryo-learningpool:
This PR adds a "Show Volume Control" and "Default Volume" setting to the media component

On mobile devices these settings have no effect and we don't try to set them.  This is because Android and iOS audio levels are controlled by the devices themselves.

screen shot 2017-06-08 at 12 18 56

### DIFF
--- a/example.json
+++ b/example.json
@@ -13,7 +13,7 @@
     "_useClosedCaptions":true,
     "_startLanguage":"en",
     "_showVolumeControl":true,
-    "_startVolume":"40%",
+    "_startVolume":"80%",
     "_playsinline": false,
     "_media": {
         "mp4": "course/assets/big_buck_bunny.mp4",
--- a/example.json
+++ b/example.json
@@ -12,6 +12,8 @@
     "_setCompletionOn":"inview",
     "_useClosedCaptions":true,
     "_startLanguage":"en",
+    "_showVolumeControl":true,
+    "_startVolume":"40%",
     "_playsinline": false,
     "_media": {
         "mp4": "course/assets/big_buck_bunny.mp4",
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -6,9 +6,8 @@ define([
 ], function(Adapt, ComponentView) {
 
     var froogaloopAdded = false;
-    var isMobile = mejs.MediaFeatures.isiOS || mejs.MediaFeatures.isAndroid;
 
-    // The following function is used to to prevent a memory leak in Internet Explorer 
+    // The following function is used to to prevent a memory leak in Internet Explorer
     // See: http://javascript.crockford.com/memory/leak.html
     function purge(d) {
         var a = d.attributes, i, l, n;
@@ -367,19 +366,9 @@ define([
 
             this.addThirdPartyAfterFixes();
 
-            if (!isMobile) {
-                this.setInitialVolume();
-            }
-
             this.setReadyStatus();
             this.setupEventListeners();
         },
-
-         setInitialVolume: function() {
-             var volume = this.model.get('_startVolume') === undefined ? '30%' : this.model.get('_startVolume');
-
-             this.mediaElement.player.setVolume(parseInt(volume)/100); 
-         },
 
         addThirdPartyAfterFixes: function() {
             var media = this.model.get("_media");
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -87,6 +87,10 @@ define([
                 }
             }
 
+            if(this.model.has('_startVolume')) {
+                modelOptions.startVolume = parseInt(this.model.get('_startVolume'))/100;
+            }
+
             modelOptions.success = _.bind(this.onPlayerReady, this);
 
             if (this.model.get('_useClosedCaptions')) {
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -87,10 +87,6 @@ define([
                 }
             }
 
-            if(this.model.has('_startVolume')) {
-                modelOptions.startVolume = parseInt(this.model.get('_startVolume'))/100;
-            }
-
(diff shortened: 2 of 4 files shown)


properties.schema
```diff
@@ -163,6 +163,25 @@
...
+      "title": "Show Volume Control (Not supported in mobile devices)",
+      "inputType": {"type": "Boolean", "options": [true, false]},
+      "validators": [],
+      "help": "If set to 'true', the volume control will appear in the media player. On mobile devices the audio level is under the user's physical control."
+    },
+    "_startVolume": {
+      "type":"string",
+      "required":true,
+      "enum": ["0%", "10%", "20%", "30%", "40%", "50%", "60%", "70%", "80%", "90%", "100%"],
+      "default": "30%",
```
moloko: 30% as the default seems very low...! I think mediaelement defaults to 80 so should this not match that?

example.json
```diff
@@ -12,6 +12,8 @@
     "_setCompletionOn":"inview",
     "_useClosedCaptions":true,
     "_startLanguage":"en",
+    "_showVolumeControl":true,
+    "_startVolume":"40%",
```
moloko: framework users do tend to copy-paste from this file so I would set this to either 80% or 100%

properties.schema
barryo-learningpool: I'm happy to change it to 80%.  In my testing I was using this clip: https://www.youtube.com/watch?v=FG1LGKieTxY which is a really loud clip so even 10% is plenty on that one.

properties.schema
```diff
@@ -163,6 +163,25 @@
...
+    "_startVolume": {
+      "type":"string",
+      "required":true,
+      "enum": ["0%", "10%", "20%", "30%", "40%", "50%", "60%", "70%", "80%", "90%", "100%"],
+      "default": "80%",
+      "title": "Default volume (Not supported on mobile devices)",
+      "inputType": {"type": "Select", "options":["0%", "10%", "20%", "30%", "40%", "50%", "60%", "70%", "80%", "90%", "100%"]},
+      "validators": ["required"],
+      "help": "Defines the default volume. On mobile devices the audio level is under the user's physical control"
+    },
```
tomgreenfield: I'd be tempted to have this an `"inputType": "number"` field instead, as that's how percentages have been done in other core plugins.

js/adapt-contrib-media.js
```diff
@@ -82,6 +83,9 @@ define([
                 if (this.model.get("_allowFullScreen") && !$("html").is(".ie9")) {
                     modelOptions.features.push('fullscreen');
                 }
+                if (!isMobile && this.model.get('_showVolumeControl')) {
```
tomgreenfield: I don't believe the `!isMobile` check is required here. The volume control doesn't show up in my testing on Android 7 and iOS 10, when included in features.

js/adapt-contrib-media.js
```diff
@@ -363,10 +367,20 @@ define([
 
             this.addThirdPartyAfterFixes();
 
+            if (!isMobile) {
```
tomgreenfield: Could use `Adapt.device.touch` instead if you fancied using the built-in attribute.

moloko: the only problem with this is that devices like windows surface will show up as 'touch'

js/adapt-contrib-media.js
barryo-learningpool: We tried the html5 video element (from https://www.w3schools.com/html/html5_video.asp) on a Nexus 5X Running Android 7.1.2 and the volume control did show up but only acted the mute/unmute the audio.

tomgreenfield: That link isn't vindicitive of Media Element's player.

It actually handles it by itself: https://github.com/mediaelement/mediaelement/blob/2.21.2/build/mediaelement-and-player.js#L3978.

barryo-learningpool: That's handy @tomgreenfield. I'll remove this check. Cheers

properties.schema
barryo-learningpool: It seems handy to have a dropdown menu. A slider would be even nicer.

js/adapt-contrib-media.js
```diff
@@ -363,10 +367,20 @@ define([
...
+                this.setInitialVolume();
+            }
+
             this.setReadyStatus();
             this.setupEventListeners();
         },
 
+        
setInitialVolume: function() {
+             var volume = this.model.get('_startVolume') === undefined ? '30%' : this.model.get('_startVolume');
```
brian-learningpool: Should this be 80% rather than 30%, for consistency with the default for `_startVolume` in `properties.schema`?

properties.schema
brian-learningpool: A percentage slider would be nice.  

I can see @tomgreenfield's point, but `number` doesn't have strict ranges now and would require more validation on in the code, i.e. checking that it isn't negative, it isn't more than 100, etc.

I think a dropdown in 10% increments for volume is a sensible offering though.

barryo-learningpool:
Thanks @moloko, I've updated the PR now

moloko:
@barryo-learningpool the code to set the start volume seems to have disappeared completely now...

barryo-learningpool:
Sorry @moloko, fixed it now 😊

barryo-learningpool:
Thanks @moloko, we need this available for IE8 so I've made the change you suggested

js/adapt-contrib-media.js
```diff
@@ -370,6 +366,10 @@ define([
 
             this.addThirdPartyAfterFixes();
 
+            if(this.model.has('_startVolume')) {
```
moloko: please could you add a comment along the lines of 'setting the start volume only works with the Flash-based player if you do it here rather than in setupPlayer'

barryo-learningpool: No problem, I've added a comment now

PR #168, opened 2017-06-22
issue/1598: update media:stop behaviour

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1598

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -180,6 +180,11 @@ define([
         },
 
         onMediaElementPlay: function(event) {
+
+            if (this.model.get("_stopOtherMediaOnPlay")) {
+                Adapt.trigger("media:stop", this);
+            }
+
             this.model.set({
                 '_isMediaPlaying': true,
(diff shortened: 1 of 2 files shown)


js/adapt-contrib-media.js
```diff
@@ -180,6 +180,11 @@ define([
         },
 
         onMediaElementPlay: function(event) {
+
+            if (this.model.get("_stopOtherMediaOnPlay")) {
```
moloko: Do we need this property? just thinking it could be the responsibility of other media plugins to decide whether they should stop when other media played or not. that way I could have a plugin that could be configured to either ignore the stop event or not, as required.

oliverfoster: I'll ask Nick in the chat. See what his use-case was.

oliverfoster: He's happy. I'll take it out.

moloko:
Note to self (or whoever ends up doing the release that includes this) - the `media:stop` event needs to be documented in README.md

PR #169, opened 2017-06-27
media : removed IE8 hack

akshay-bhavsar:


moloko:
@sarveshwar-gavhane good spot! @akshay-bhavsar please remove the IE9 check as suggested

moloko:
**as with all of these, please leave unmerged for now**

oliverfoster:
does the purge function need to go?

oliverfoster:
according to the ticket the issue showed up in IE11 and was solved by Crockfords IE6/IE7 fix.... so,let's leave it for the moment. @tomgreenfield be wary of memory leaks with the next version of mejs.
here is the original issue https://github.com/adaptlearning/adapt_framework/issues/1223

PR #170, opened 2017-06-28
fixes #1622

AaronmcL:
Changed type input property to select dropdown menu

### DIFF
--- a/properties.schema
+++ b/properties.schema
@@ -91,11 +91,12 @@
         "type": {
           "type":"string",
           "required":false,
+          "enum": ["Default", "video/youtube", "video/vimeo"],
           "default": "",
           "title": "Type",
-          "inputType": "Text",
+          "inputType": {"type": "Select", "options":["Default", "video/YouTube", "video/Vimeo"]},
           "validators": [],
-          "help": "Enter the 'Source URL' type, for a YouTube link this would be video/youtube, and video/vimeo for Vimeo"
+          "help": "Select option from below"
         },
         "poster": {
           "type":"string",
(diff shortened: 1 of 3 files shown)


properties.schema
```diff
@@ -91,11 +91,12 @@
...
           "type":"string",
           "required":false,
+          "enum": ["Default", "video/youtube", "video/vimeo"],
           "default": "",
           "title": "Type",
-          "inputType": "Text",
+          "inputType": {"type": "Select", "options":["Default", "video/YouTube", "video/Vimeo"]},
           "validators": [],
-          "help": "Enter the 'Source URL' type, for a YouTube link this would be video/youtube, and video/vimeo for Vimeo"
+          "help": "Select option from below"
```
brian-learningpool: Can you change this to this? 
"Select the 'Source URL' type, for a YouTube link this would be 'video/youtube', and 'video/vimeo' for Vimeo."

properties.schema
```diff
@@ -91,11 +91,12 @@
         "type": {
           "type":"string",
           "required":false,
+          "enum": ["Default", "video/youtube", "video/vimeo"],
```
brian-learningpool: For backwards compatibility "Default" should actually be an empty string "".

properties.schema
```diff
@@ -91,11 +91,12 @@
...
           "default": "",
           "title": "Type",
-          "inputType": "Text",
+          "inputType": {"type": "Select", "options":["Default", "video/YouTube", "video/Vimeo"]},
```
brian-learningpool: As above regarding "Default", but the other array items should probably be lowercase too.

moloko:
@sarveshwar-gavhane don't worry too much about that, I think it's easier just to deal with the version number at the point of release - I'll sort it out then

PR #172, opened 2017-08-06
fix null reference on IE v11.0.9600.18738

loudoweb:
mediaElement was null on page loading on this particular IE version.

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -256,11 +256,12 @@ define([
 
             // Make sure this view isn't triggering media:stop
             if (view && view.cid === this.cid) return;
-
-            var player = this.mediaElement.player;
-            if (!player) return;
             
-            player.pause();
+            if (this.mediaElement && this.mediaElement.player) {
+              var player = this.mediaElement.player;
+              player.pause();
+            }
+            
         },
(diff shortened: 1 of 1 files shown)


js/adapt-contrib-media.js
```diff
@@ -256,11 +256,12 @@ define([
 
             // Make sure this view isn't triggering media:stop
             if (view && view.cid === this.cid) return;
-
-            var player = this.mediaElement.player;
-            if (!player) return;
             
-            player.pause();
+            if (this.mediaElement && this.mediaElement.player) {
+              var player = this.mediaElement.player;
```
taylortom: We don't really need this var any more, so can probably go with something as simple as:

_(as we like to return early)_
```
if(!this.mediaElement || !this.mediaElement.player) return;

this.mediaElement.player.pause();
```
or
```
if(this.mediaElement && this.mediaElement.player) {
    this.mediaElement.player.pause();
}
```

moloko:
@loudoweb is there a related issue for this? sorry but we require a matching issue for all PRs [so that we can keep track of everything](https://github.com/adaptlearning/adapt_framework/wiki/Contributing-code#before-submitting)

PR #173, opened 2017-08-11
mejs-overlay hover state bug fix

guywillis:
Override the default hover state for 'mejs-overlay' and tie it to Modernizr's 'no-touch' class to stop the hover state from appearing on touch devices.

### DIFF
--- a/less/media.less
+++ b/less/media.less
@@ -80,6 +80,23 @@
         width:100%;
         display:block;
     }
+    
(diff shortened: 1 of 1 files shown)


moloko:
@guywillis we do require a matching issue describing the problem though

guywillis:
https://github.com/adaptlearning/adapt_framework/issues/1714

PR #176, opened 2018-01-18
Avoid use of relative protocol when loading remote media

mediabounds:
By forcing remote media to use a relative protocol, it causes the content to fail to load if it is running over a protocol other than `http:` or `https:` (e.g. `file:`). (Additionally, the use of relative protocols is considered an anti-pattern by some: https://www.paulirish.com/2010/the-protocol-relative-url/).

This only changes the source if the page is being loaded over a secure connection and the remote content is not; to avoid accidentally loading Mixed Content, it attempts to load the asset over https instead.

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -53,13 +53,14 @@ define([
             });
 
             if (this.model.get('_media').source) {
-                // Remove the protocol for streaming service.
-                // This prevents conflicts with HTTP/HTTPS
                 var media = this.model.get('_media');
 
-                media.source = media.source.replace(/^https?\:/, "");
+                // Avoid loading of Mixed Content (insecure content on a secure page)
+                if (window.location.protocol === 'https:' && media.source.indexOf('http:') === 0) {
+                    media.source = media.source.replace(/^http\:/, 'https:');
+                }
 
-                this.model.set('_media', media); 
+                this.model.set('_media', media);
             }
 
             this.checkIfResetOnRevisit();


moloko:
> Additionally, the use of relative protocols is considered an anti-pattern by some: https://www.paulirish.com/2010/the-protocol-relative-url/
Clearly Mr Irish has not had a lot of experience producing content (like e-learning) where it cannot generally be known in advance whether the server that hosts that content will use http or https ;-)

Would you mind [logging an issue](https://github.com/adaptlearning/adapt_framework/issues/new) for this? Having a matching issue for every change helps us to keep track

js/adapt-contrib-media.js
```diff
@@ -53,13 +53,14 @@ define([
...
 
             if (this.model.get('_media').source) {
-                // Remove the protocol for streaming service.
-                // This prevents conflicts with HTTP/HTTPS
                 var media = this.model.get('_media');
 
-                media.source = media.source.replace(/^https?\:/, "");
+                // Avoid loading of Mixed Content (insecure content on a secure page)
+                if (window.location.protocol === 'https:' && media.source.indexOf('http:') === 0) {
+                    media.source = media.source.replace(/^http\:/, 'https:');
```
oliverfoster: Https would be implied here, so you could just drop it.
```js
media.source = media.source.replace(/^http\:/, ''); 
```

mediabounds:
Issue opened: https://github.com/adaptlearning/adapt_framework/issues/1926

js/adapt-contrib-media.js
mediabounds: My thinking and intention here was to be explicit with the protocols. But, the end result would be the same either way, so I'm open to changing it.

moloko: Although @oliverfoster is quite right I think in this instance the way it is currently written makes the intention of the code clearer

PR #178, opened 2018-02-05
add aria-expanded attribute to the transcript button

moloko:
fixes https://github.com/adaptlearning/adapt_framework/issues/1956


moloko:
@oliverfoster is it still correct for this to be using `$(window).resize();` or should we switch to one of the built-in Adapt functions now?

oliverfoster:
I don't think it's needed. Trickle should manage its own height. It would be worth checking without it.

moloko:
> I don't think it's needed. Trickle should manage its own height. 

It's more when media is used inside things like ABS I think. Happy to leave as-is, just thought it was a good opportunity to ask the question.

oliverfoster:
need changing to aria-pressed https://github.com/adaptlearning/adapt_framework/issues/2134

moloko:
I think this is probably now superseded by #183 - we should focus on getting that in then if there's anything still needed here I'll update it

oliverfoster:
sorry, I didn't mean to duplicate the work. should've checked.

moloko:
TBH it's good that you did cause mine isn't quite right following my ham-fisted attempt to sort it out following on from Simon's fix to how the button label text was applied!

moloko:
superseded by #183

PR #181, opened 2018-03-26
issue/#2024: Video controls disappear in iOS 11

boonier:
fixes https://github.com/adaptlearning/adapt_framework/issues/2024

### DIFF
--- a/less/media.less
+++ b/less/media.less
@@ -152,6 +152,12 @@
         background-color: black;
     }
 
+    // Bug fix for missing control icons on iOS touch devices 
+    // ==============================
+    .mejs-offscreen {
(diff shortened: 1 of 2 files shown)


tomgreenfield:
Thanks for the fix, @boonier!

Would you mind moving the style to our override file, [`media.less`](https://github.com/adaptlearning/adapt-contrib-media/blob/master/less/media.less), so that `mep.less` remains a carbon copy of Media Element's CSS?

boonier:
This is done @tomgreenfield

moloko:
I do get this warning pop up in Visual Studio Code - can it safely be ignored? 
image

moloko:
sorry ignore me - of course it can be ignored because this is extending (rather than overriding) the `.mejs-offscreen` class from mep.less... 

as you were!

PR #182, opened 2018-05-04
Updated to only change the button text

simondate:
The Media component currently changes the DOM structure when the transcript button is opened or closed. This change makes the `transcript-text-container` be updated instead which keeps the DOM structure the same.

resolves https://github.com/adaptlearning/adapt_framework/issues/2062

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -399,20 +399,20 @@ define([
         onToggleInlineTranscript: function(event) {
             if (event) event.preventDefault();
             var $transcriptBodyContainer = this.$(".media-inline-transcript-body-container");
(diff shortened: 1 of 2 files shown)


moloko:
Makes me think: do we even need the text to be inside a `.transcript-text-container` div - couldn't it just be:
```html
<button class="media-inline-transcript-button button">Transcript</button>
```
As per Submit, Show Feedback etc.?

(While you're there - any chance you could remove `role="button"` from the template?)

And would you mind [logging an issue](https://github.com/adaptlearning/adapt_framework/issues/new) for this?

simondate:
I would prefer to keep `.transcript-text-container` as we've had a few project we're we've added icons inside of the button.

capture

I'll create an issue now.

PR #185, opened 2018-11-05
Poster image with audio

simondate:
Makes sense as a feature. I remember we did this at Kineo. Double check for both the mp3 file and for the poster so nothing will change unless the user is searching for it

[#2264](https://github.com/adaptlearning/adapt_framework/issues/2264)

### DIFF
--- a/templates/media.hbs
+++ b/templates/media.hbs
@@ -5,6 +5,12 @@
     <button class="aria-label js-skip-to-transcript" aria-label="{{_globals._components._media.skipToTranscript}}"></button>
     {{/any}}
     <div class="media-widget component-widget a11y-ignore-aria">
+    
+    {{#if _media.mp3}}
+      {{#if _media.poster}}
+        <img src="{{_media.poster}}" style="width: 100%; height: 100%;"/>
+      {{/if}}
+    {{/if}}
 
     {{#if _media.mp3}}
       <audio src="{{_media.mp3}}" type="audio/mp3" style="width: 100%; height: 100%;"/>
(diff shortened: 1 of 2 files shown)


templates/media.hbs
```diff
@@ -5,6 +5,12 @@
     <button class="aria-label js-skip-to-transcript" aria-label="{{_globals._components._media.skipToTranscript}}"></button>
     {{/any}}
     <div class="media-widget component-widget a11y-ignore-aria">
+    
+    {{#if _media.mp3}}
+      {{#if _media.poster}}
```
moloko: this line and the following two could be moved inside the existing `{{#if _media.mp3}}` on L15

lc-thomasberger:
Is there a related issue?

templates/media.hbs
simondate: Haha yes, that would be simpler. Sorry I was working quite late and didn't really what I was doing.

simondate:
@lc-thomasberger I can create one if it's required. I think it's quite self-explanatory though?

oliverfoster:
Issue > PR is the standard workflow at the moment. It helps us keep track of what's needed and what is going on.

moloko:
https://github.com/adaptlearning/adapt_framework/wiki/Contributing-code#before-submitting

simondate:
https://github.com/adaptlearning/adapt_framework/issues/2264

oliverfoster:
ta

brian-learningpool:
Indentation was already a bit inconsistent in the .hbs, sometimes 2 sometimes 4 spaces.  I don't think @simondate's PR has introduced anything in regard to this?  Are we ok to merge this and address the indentation inconsistencies as a separate housekeeping task?

PR #186, opened 2018-11-12
Fix for issue/#2266

chris-steele:


js/adapt-contrib-media.js
```diff
@@ -109,7 +109,13 @@ define([
                 }
             }
 
-            modelOptions.success = _.bind(this.onPlayerReady, this);
+            /*
+                Unless we are on Android/iOS and using native controls, when MediaElementJS initializes the player it will invoke the success callback prior to performing one last call to setPlayerSize. This call to setPlayerSize is deferred by 50ms so we add a delay of 100ms here to ensure that we don't invoke setReadyStatus until the player is definitely finished rendering.
+            */
+
+            var successCallback = _.bind(this.onPlayerReady, this);
```
moloko: might as well switch to native `Function.bind` while you're here

js/adapt-contrib-media.js
```diff
@@ -109,7 +109,13 @@ define([
...
             }
 
-            modelOptions.success = _.bind(this.onPlayerReady, this);
+            /*
+                Unless we are on Android/iOS and using native controls, when MediaElementJS initializes the player it will invoke the success callback prior to performing one last call to setPlayerSize. This call to setPlayerSize is deferred by 50ms so we add a delay of 100ms here to ensure that we don't invoke setReadyStatus until the player is definitely finished rendering.
+            */
+
+            var successCallback = this.onPlayerReady.bind(this);
+
+            modelOptions.success = function(mediaElement, domObject) {_.delay(function(){successCallback(mediaElement, domObject)}, 100);};
```
brian-learningpool: Can you add some line-breaks to this please?

chris-steele: It did look quite the headache - sorted now

js/adapt-contrib-media.js
```diff
@@ -109,7 +109,17 @@ define([
...
 
-            modelOptions.success = _.bind(this.onPlayerReady, this);
+            /*
+                Unless we are on Android/iOS and using native controls, when MediaElementJS initializes the player it will invoke the success callback prior to performing one last call to setPlayerSize. This call to setPlayerSize is deferred by 50ms so we add a delay of 100ms here to ensure that we don't invoke setReadyStatus until the player is definitely finished rendering.
+            */
+
+            var successCallback = this.onPlayerReady.bind(this);
+
+            modelOptions.success = function(mediaElement, domObject) {
+                _.delay(function() {
```
moloko: rubbish timing I know but just to point out you could do this instead
```suggestion
                _.delay(successCallback, 100, mediaElement, domObject);
```

oliverfoster: ^ just for reference https://underscorejs.org/#delay

chris-steele: I'm all for the succinct , thanks :+1:

js/adapt-contrib-media.js
```diff
@@ -109,7 +109,15 @@ define([
...
             }
 
-            modelOptions.success = _.bind(this.onPlayerReady, this);
+            /*
+                Unless we are on Android/iOS and using native controls, when MediaElementJS initializes the player it will invoke the success callback prior to performing one last call to setPlayerSize. This call to setPlayerSize is deferred by 50ms so we add a delay of 100ms here to ensure that we don't invoke setReadyStatus until the player is definitely finished rendering.
+            */
+
+            var successCallback = this.onPlayerReady.bind(this);
+
+            modelOptions.success = function(mediaElement, domObject) {
```
oliverfoster:
```js
modelOptions.success = _.debounce(this.onPlayerReady.bind(this), 100);
```
Would have the same effect but with slightly less clarity. Leave it as is. T'is fine.

chris-steele: Yeah I like this, the arguments get passed automatically (@moloko's request) and it doesn't look awkward on one line (@brian-learningpool's request). The comment above gives it further clarity so I think it's worth the change.

PR #188, opened 2018-11-30
have media component remember learner's choice of captions

moloko:
see https://github.com/adaptlearning/adapt_framework/issues/2284

Well, this took a lot longer than I thought as I'd completely forgotten that the media component supports multiple caption languages!

On the off chance that media component 'a' has a captions language that's not present in media component 'b' - and the user selects that language, 'b' will default to whatever language is set in `_startLanguage`. If that's not set, it will simply turn its own captions off. The actual language the user chose will be the one stored in `offlineStorage`.

also fixes https://github.com/adaptlearning/adapt_framework/issues/1883 - would be interested to get some comments on how I have implemented this (using `className` function in adapt-contrib-media.js) - the alternative would be to do this:
toggle-captions-media-hbs
which is less code - but it does seem that a lot of themes still have copies of the core templates, so those wouldn't pick up this change. Happy to switch if people think doing it via media.hbs is better.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "adapt-contrib-media",
   "version": "3.0.2",
-  "framework": ">=3",
+  "framework": ">=3.1",
   "homepage": "https://github.com/adaptlearning/adapt-contrib-media",
   "issues": "https://github.com/adaptlearning/adapt_framework/issues/new",
   "displayName": "Media",
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -124,11 +124,14 @@ define([
             This call to setPlayerSize is deferred by 50ms so we add a delay of 100ms here to ensure that
             we don't invoke setReadyStatus until the player is definitely finished rendering.
             */
-
             modelOptions.success = _.debounce(this.onPlayerReady.bind(this), 100);
 
             if (this.model.get('_useClosedCaptions')) {
-                modelOptions.startLanguage = this.model.get('_startLanguage') === undefined ? 'en' : this.model.get('_startLanguage');
+                var startLanguage = this.model.get('_startLanguage') || 'en';
+                if (!Adapt.offlineStorage.get('captions')) {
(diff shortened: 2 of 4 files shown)


moloko:
> In case if I have two or more media component on the same page it doesn't give proper result.

@NayanKhedkar could you elaborate on that? 

> I think the condition must be update to `if (view && view.cid !== this.cid) return;`.

That would mean the media component instance would only respond to 'captions change' events broadcast by itself which is the exact opposite of what I want it to do

> this line set same caption language for all the media components on the page

that's [what I'm trying to achieve with this change](https://github.com/adaptlearning/adapt_framework/issues/2284).

NayanKhedkar:
> > In case if I have two or more media component on the same page it doesn't give proper result.
> 
> @NayanKhedkar could you elaborate on that?
> 
> > I think the condition must be update to `if (view && view.cid !== this.cid) return;`.
> 
> That would mean the media component instance would only respond to 'captions change' events broadcast by itself which is the exact opposite of what I want it to do
> 
> > this line set same caption language for all the media components on the page
> 
> that's [what I'm trying to achieve with this change](https://github.com/adaptlearning/adapt_framework/issues/2284).



I have added two media component.
here is scenario.
Select the none caption from first media component,Then select english caption from second media component.You found that the first media component cc changed to english.

you can find here
```if (view && view.cid === this.cid) return;```
https://nayankhedkar.github.io/Test/build01/#/id/co-05

```if (view && view.cid !== this.cid) return;```
https://nayankhedkar.github.io/Test/build02/#/id/co-05

moloko:
> You found that the first media component cc changed to english.

Yes, that's exactly the behaviour that I want! Otherwise, as a learner, if there are 20 media components in the course I have to switch on captions 20 times... this PR fixes that by applying the user's choice throughout all the media components in the course, just like the media component already does for the volume setting.

moloko:
@NayanKhedkar this is the feature request that describes what I'm trying to do https://github.com/adaptlearning/adapt_framework/issues/2284

NayanKhedkar:
Okay got it.

PR #190, opened 2019-02-01
#2344 add global triggers for play, pause and ended events

danielstorey:
[#2344](https://github.com/adaptlearning/adapt_framework/issues/2344)

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -518,9 +518,12 @@ define([
         },
 
         triggerGlobalEvent: function(eventType) {
-            var mediaType = this.mediaElement.tagName.toLowerCase();
-            var event = [mediaType, eventType].join(':');
-            Adapt.trigger(event, this.mediaElement.src, this.mediaElement.pluginType);
+            Adapt.trigger('media', {
+              isVideo: this.mediaElement.player.isVideo,
+              type: eventType,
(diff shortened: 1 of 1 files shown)


js/adapt-contrib-media.js
```diff
@@ -510,6 +515,12 @@ define([
             if (this.model.get('_transcript')._setCompletionOnView !== false) {
                 this.setCompletionStatus();
             }
+        },
+
+        triggerGlobalEvent: function(eventType) {
+            var mediaType = this.mediaElement.tagName.toLowerCase();
+            var event = [mediaType, eventType].join(':');
+            Adapt.trigger(event, this.mediaElement.src, this.mediaElement.pluginType);
```
oliverfoster: I'd go for:
```js
Adapt.trigger("media:"+eventType, this.mediaElement);
```
Rather than multiplying the events based on tagname and event type.
It passes a lot more information through this way.

danielstorey: That's a fair point, although I anticipate adding these events to other extensions as well, which may not necessarily have a normalised mediaElement

oliverfoster: Cool.
```js
Adapt.trigger("media:"+eventType, {
  src: this.mediaElement.src,
  // etc
});
```
^ This is easier to extend

PR #192, opened 2019-02-21
Resolves #2357 debounce video events

danielstorey:
Queue a media event first to see if it should be fired. This:
- Prevents pause firing when ended
- Prevents play/pause firing simultaneously when user seeks

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -282,7 +282,7 @@ define([
         },
 
         onMediaElementPlay: function(event) {
-            this.triggerGlobalEvent('play');
+            this.queueGlobalEvent('play');
 
             Adapt.trigger("media:stop", this);
 
@@ -297,13 +297,13 @@ define([
         },
 
         onMediaElementPause: function(event) {
-            this.triggerGlobalEvent('pause');
+            this.queueGlobalEvent('pause');
 
             this.model.set('_isMediaPlaying', false);
         },
 
         onMediaElementEnded: function(event) {
-            this.triggerGlobalEvent('ended');
+            this.queueGlobalEvent('ended');
 
             this.model.set('_isMediaEnded', true);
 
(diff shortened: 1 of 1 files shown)


moloko:
Sorry if I'm missing something here but couldn't you just use `_.debounce`?

danielstorey:
> Sorry if I'm missing something here but couldn't you just use `_.debounce`?

@moloko it's a fair point and I thought someone might mention that - I should have left a comment!

For 'ended' events you'd have to guarantee that 'pause' and 'ended' fire in the same order each time and I wouldn't want to take that risk. It would only take something like IE to come along and ruin it.

Also debouncing play/pause when seeking would mean that one of them would still fire. This function prevents either firing as there would be no merit in doing so.

moloko:
> It would only take something like IE to come along and ruin it.

😆

js/adapt-contrib-media.js
```diff
@@ -517,6 +517,36 @@ define([
...
+            clearTimeout(this.eventTimeout);
+
+            // Always trigger 'ended' events
+            if (eventType === 'ended') {
+                return this.triggerGlobalEvent(eventType);
+            }
+
+            // Fire the event after a delay, only if another event has not just been fired
+            if (timeSinceLastEvent > debounceTime) {
+                this.eventTimeout = setTimeout(this.triggerGlobalEvent.bind(this, eventType), debounceTime);
```
oliverfoster: if someone double clicks the play button within the 0.5s gap it will never trigger a pause event, just a play event with no ended or paused event.

are you sure this is useful? i know the events can fire in quick succession but dropping events arbitrarily within 0.5s seems like adding wood to the fire?

oliverfoster: this code is deeply confusing me.
1. clear waiting timeouts
2. if ended trigger event, return
3. drop event if under 500ms since last event trigger, return (< which events are going to get dropped and why are the earlier ones taking precedent?)
4. queue event, wait 500ms, trigger last event type

danielstorey: I'm happy to discuss a better solution but yes if events occur within a 500ms window of another event they will get dropped. This is to prevent play/pause firing in quick succession and cancelling each other out, e.g. when the user seeks. Potentially the odd play event will get through if the user is determined to mess things up by scrubbing loads but
1. This is unlikely
2. Still better than loads of play/pause events firing without fail

The ended event is always significant so should always be allowed to fire.

The use case I am basing this work on is regarding analytics. There's no specific functionality that requires these events to be 100% reliable.

Like I say though, I am open to other suggestions.

oliverfoster: It seems fine now that I get it. Whether the media component is the right place to filter these events would depend on future uses, so we'll have to see if the filter needs moving at some point later.

Being a bit more wordy on the variable names would've been cool and a return early at the end to avoid nesting and make the primary intent clearer.

Sorry for getting befuddled. 
```js
        filterRapidEvents: function(eventType) {
            var rapidFireTimeoutWindow = 500;

            var currentTime = Date.now();
            var lastEventTime = this.lastEventTime || 0;
            this.lastEventTime = currentTime;

            clearTimeout(this.waitingTriggerTimeout);

            var isAlwaysTriggered = (eventType === 'ended')
            if (isAlwaysTriggered) {
                this.triggerGlobalEvent(eventType);
                return;
            }

            var lastEventDuration = (currentTime - lastEventTime);
            var isRapidlyFiring = (lastEventDuration <= rapidFireTimeoutWindow );
            if (isRapidlyFiring) return;

            // Wait for rapidFireTimeoutWindow until triggering
            // Allows trigger to be dropped if another incoming event arrives
            var trigger = this.triggerGlobalEvent.bind(this, eventType);
            this.waitingTriggerTimeout = setTimeout(trigger, rapidFireTimeoutWindow);

        },
```

PR #194, opened 2019-04-08
#2128 trigger pause when media component leaves view

jamesrea83:
In response to [#2128](https://github.com/adaptlearning/adapt_framework/issues/2128).

As per [this comment](https://github.com/adaptlearning/adapt_framework/issues/2128#issuecomment-396867757) media component is paused directly, rather than triggering `media:stop`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ guide the learner’s interaction with the component.
 
 **_allowFullScreen** (boolean): Determines whether fullscreen mode is available or not. Note that changing this setting has no effect in Internet Explorer 9 as this browser does not support fullscreen mode for HTML video.
 
-**_pauseWhenNotInview**  (boolean): If set to true, pause playback when video is no longer in view. The default is `false`.
+**_pauseWhenOffScreen**  (boolean): If set to true, pause playback when video is no longer in view. The default is `false`.
 
 **_playsinline** (boolean): If set to `true`, videos will play 'inline' on iPhones (the same way they do on iPads). Note that this feature is only available in iOS10 and above. The default is `false`.    
 
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ guide the learner’s interaction with the component.
 
 **_allowFullScreen** (boolean): Determines whether fullscreen mode is available or not. Note that changing this setting has no effect in Internet Explorer 9 as this browser does not support fullscreen mode for HTML video.
 
+**_pauseWhenNotInview**  (boolean): If set to true, pause playback when video is no longer in view. The default is `false`.
+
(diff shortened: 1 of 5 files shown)


moloko:
I think this behaviour should be configurable via a property. Question is: should it be enabled by default or not? Thoughts?

brian-learningpool:
I agree this setting should be configurable via a property.  

I was initially going to say defaulted to on, but I've changed my mind and gone with a default of off for two reasons:
- backwards compatibility (maintaining existing functionality)
- the Media component also handles playing MP3 files, where it's probably acceptable to continue playing while not in view

moloko:
fine by me, tbh it’s not really an issue in the framework where this sort of thing is easily changed via find-and-replace, it’s more an issue for authoring tool users.

README.md
```diff
@@ -50,6 +50,8 @@ guide the learner’s interaction with the component.
 
 **_allowFullScreen** (boolean): Determines whether fullscreen mode is available or not. Note that changing this setting has no effect in Internet Explorer 9 as this browser does not support fullscreen mode for HTML video.
 
+**_pauseWhenNotInview**  (boolean): If set to true, pause playback when video is no longer in view. The default is `false`.
```
moloko: `_pauseWhenOffScreen` might be a more meaningful name to anyone unaware of the 'inview' library

bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "adapt-contrib-media",
-  "version": "4.0.1",
+  "version": "4.0.2",
```
moloko: this adds a new feature so version number should be 4.1.0

jamesrea83:
@oliverfoster what do you think about attaching the listener only when playback starts, and removing it when it stops? As opposed to the current adding it to setupEventListeners / remove.

oliverfoster:
> @oliverfoster what do you think about attaching the listener only when playback starts, and removing it when it stops? As opposed to the current adding it to setupEventListeners / remove.

It would be preferable from a CPU usage perspective to attached the inview listeners only when they are needed but could also make for more sprawling code. Having good names and layouts of the functions would mitigate the secondary layer of event attachments / removals and you will still need to remove the events when the component is destroyed.

js/adapt-contrib-media.js
```diff
@@ -286,6 +286,9 @@ define([
 
             Adapt.trigger("media:stop", this);
 
+            var pauseWhenOffScreen = this.model.get('_pauseWhenOffScreen');
+            if (pauseWhenOffScreen) $(this.mediaElement).on('inview', this.onMediaElementInview);
```
moloko: minor thing I know, but the variable `pauseWhenOffScreen` is a bit unnecessary since it's only referenced once

PR #196, opened 2019-04-29
Fix: stuck in infinite loading loop when no media is selected

canstudios-simonasd:
fixes https://github.com/adaptlearning/adapt_framework/issues/2418

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -150,21 +150,15 @@ define([
                 this.cleanUpPlayer();
 
                 var _media = this.model.get('_media');
-                // if there's no _media property, set ready now, as the success callback will never be called
-                if (!_media) {
-                    Adapt.log.warn("ERROR! No _media property found in components.json for component " + this.model.get('_id'));
+                // if no media is selected - set ready now, as success won't be called
+                if (!_media.mp3 && !_media.mp4 && !_media.ogv && !_media.webm && !_media.source) {
+                    Adapt.log.warn("ERROR! No media is selected in components.json for component " + this.model.get('_id'));
                     this.setReadyStatus();
                     return;
                 }
                 // Check if we're streaming
                 if (_media.source) {
                     this.$('.media-widget').addClass('external-source');
-                    return;
-                }
-                // if no media is selected - set ready now, as success won't be called
-                if (!_media.mp3 && !_media.mp4 && !_media.ogv && !_media.webm) {
-                    Adapt.log.warn("ERROR! No media is selected in components.json for component " + this.model.get('_id'));
-                    this.setReadyStatus();
                 }
             }.bind(this));
         },
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -149,14 +149,23 @@ define([
                 this.$('audio, video').mediaelementplayer(modelOptions);
                 this.cleanUpPlayer();
 
-                // We're streaming - set ready now, as success won't be called above
-                try {
-                    if (this.model.get('_media').source) {
-                        this.$('.media-widget').addClass('external-source');
-                    }
-                } catch (e) {
+
+                var _media = this.model.get('_media');
+                // Check if _media property exists - set ready now, as success won't be called
+                if (!_media) {
                     console.log("ERROR! No _media property found in components.json for component " + this.model.get('_id'));
                     this.setReadyStatus();
+                } else {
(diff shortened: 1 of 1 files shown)


js/adapt-contrib-media.js
```diff
@@ -149,14 +149,23 @@ define([
                 this.$('audio, video').mediaelementplayer(modelOptions);
                 this.cleanUpPlayer();
 
-                // We're streaming - set ready now, as success won't be called above
-                try {
-                    if (this.model.get('_media').source) {
-                        this.$('.media-widget').addClass('external-source');
-                    }
-                } catch (e) {
+
```
moloko: delete empty line

js/adapt-contrib-media.js
```diff
@@ -149,14 +149,23 @@ define([
...
+                var _media = this.model.get('_media');
+                // Check if _media property exists - set ready now, as success won't be called
```
moloko:
```suggestion
                //if there's no _media property, set ready now, as the success callback will never be called
```

js/adapt-contrib-media.js
```diff
@@ -149,14 +149,23 @@ define([
...
+                if (!_media) {
                     console.log("ERROR! No _media property found in components.json for component " + this.model.get('_id'));
                     this.setReadyStatus();
+                } else {
```
moloko: `return` here to make the `else` unnecessary

js/adapt-contrib-media.js
```diff
@@ -149,14 +149,23 @@ define([
...
+                    // Check if we're streaming
+                    if (_media.source) {
+                        this.$('.media-widget').addClass('external-source');
+                    } else {
```
moloko: `return` here to make the `else` unnecessary

js/adapt-contrib-media.js
```diff
@@ -149,14 +149,23 @@ define([
...
+                        // If no media is selected - set ready now, as success won't be called
+                        if (!_media.mp3 && !_media.mp4 && !_media.ogv && !_media.webm) {
+                            console.log("ERROR! No media is selected in components.json for component " + this.model.get('_id'));
```
moloko: Change from `console.log` to `Adapt.log.warn` (same goes for the one on L156)

js/adapt-contrib-media.js
```diff
@@ -149,23 +149,22 @@ define([
                 this.$('audio, video').mediaelementplayer(modelOptions);
                 this.cleanUpPlayer();
 
-
                 var _media = this.model.get('_media');
-                // Check if _media property exists - set ready now, as success won't be called
+                // if there's no _media property, set ready now, as the success callback will never be called
                 if (!_media) {
-                    console.log("ERROR! No _media property found in components.json for component " + this.model.get('_id'));
+                    Adapt.log.warn("ERROR! No _media property found in components.json for component " + this.model.get('_id'));
```
moloko: I'm wondering if we should drop this check for the `_media` object not existing at all. Just checking here isn't sufficient: if you try this without `_media` being defined in the JSON you'll see that checks also need to be added to and `preRender` and `checkForSupportedCCLanguage` - and I'm not entirely sure it's ever going to be missing in the first place as it's set as a 'required' property in the schema.

js/adapt-contrib-media.js
```diff
@@ -149,23 +149,22 @@ define([
...
+                    this.setReadyStatus();
+                    return;
+                }
+                // Check if we're streaming
+                if (_media.source) {
+                    this.$('.media-widget').addClass('external-source');
+                    return;
+                }
+                // if no media is selected - set ready now, as success won't be called
```
moloko: Sorry, I should have caught this last time, but I think these two checks should be swapped around, like this:
```js
// if no media is selected - set ready now, as success won't be called
if (!_media.mp3 && !_media.mp4 && !_media.ogv && !_media.webm) {
    Adapt.log.warn("ERROR! No media is selected in components.json for component " + this.model.get('_id'));
    this.setReadyStatus();
    return;
}
// Check if we're streaming
if (_media.source) {
    this.$('.media-widget').addClass('external-source');
}
```
Just in terms of style it's good to get all your 'check for things not being set' conditions out of the way first

canstudios-simonasd: I've added `_media.source` to first if statement, it means that first statement now checks all available media formats and second just checks if it external source or not.

PR #198, opened 2019-05-14
Ensure aria-valuemax is set to a valid value

SteveX:
fixes [#2431](https://github.com/adaptlearning/adapt_framework/issues/2431)

aria-valuemax="NaN" causes an error in a Lighthouse accessibility audit, this fixes by replacing with 0 post-render.


SteveX:
Framework issue:
https://github.com/adaptlearning/adapt_framework/issues/2431

PR #201, opened 2019-07-16
Amends to allow iOS native controls within mediaElement

danielghost:
See adaptlearning/adapt_framework#2477.


moloko:
For the benefit of anyone else who wants to test this, you need to include the following in your media component's json in order to trigger the issue:
```json
"_startVolume": "80%",
"_playerOptions": {
    "iPadUseNativeControls": true,
    "iPhoneUseNativeControls": true
}
```

PR #205, opened 2019-12-03
remove vendor prefixes where no longer required

moloko:
https://github.com/adaptlearning/adapt_framework/issues/2557

### DIFF
--- a/libraries/mediaelement-fullscreen-hook.js
+++ b/libraries/mediaelement-fullscreen-hook.js
@@ -7,49 +7,46 @@ define([
     $.extend(mejs.MediaElementPlayer.prototype, {
         /**
          * fixes a bug (adaptlearning/adapt_framework#1478)
-         * where the media player going into/coming out of full-screen mode would trigger inview on 
-         * components below it; we therefore need to switch off inview when entering full screen mode 
+         * where the media player going into/coming out of full-screen mode would trigger inview on
+         * components below it; we therefore need to switch off inview when entering full screen mode
          * and switch it back on again after exiting full screen mode
          */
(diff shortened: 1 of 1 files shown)


moloko:
To see changes without whitespace/indentation differences, go [here](https://github.com/adaptlearning/adapt-contrib-media/pull/205/files?utf8=%E2%9C%93&diff=split&w=1)

libraries/mediaelement-fullscreen-hook.js
```diff
@@ -1,55 +1,52 @@
...
+          }
+        };
+      }
+      return mepPrototype.detectFullscreenMode.apply(this, arguments);
+    },
+
+    /**
+    * because the fullscreen events and properties are still vendor-prefixed in some browsers...
+    */
+    getVendorPrefix: function() {
```
tomgreenfield: Code style suggestion:
```js
getVendorPrefix: function() {
  switch (Adapt.device.browser) {
    case 'internet explorer':
      return 'ms';
    case 'microsoft edge':
    case 'safari':
      return 'webkit';
    default:
      return '';
  }
}
```

moloko: I did consider changing to `switch` but didn't feel it really added that much. Furthermore, assuming that Safari and Edge will stop requiring a vendor prefix, it'll be less work to change.

tomgreenfield: Up to you at the end of the day!

I'd still remove the `.toLowerCase()` (looks like it's been lowercase ever since framework v2) and also fix the `prefiix` typo in the comment.

PR #208, opened 2020-07-01
Fix Microsoft edge bug

paspiz85:
Option "Attempt to prevent media scrubbing?" doesn't work on Microsoft Edge.
**Description in details**
With this this option enabled, if I seek on Microsoft Edge the variable event.target.currentTime is update immediatly (skipping the controls in onMediaElementSeeking) and _maxViewed is always updated. 
**Solution**
To prevent this behaviour I check that _maxViewed is updated only with natural video play, not with seeking

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -353,7 +353,7 @@ define([
           maxViewed = 0;
         }
         if (event.target.currentTime > maxViewed) {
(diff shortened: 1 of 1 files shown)


paspiz85:
bug is fixed on last microsoft edge update

oliverfoster:
thanks for confirming

PR #212, opened 2021-01-19
clean up eslint issues

moloko:
There were so many (355!) eslint issues that just viewing this file was hurting my eyes, with so much error highlighting! All should now be fixed.

When reviewing: please note, the aim of this was to do just enough to clear the eslint errors. I'm sure you will spot plenty of ways this could be further improved but really I would prefer to do that as a separate job!

### DIFF
--- a/js/adapt-contrib-media.js
+++ b/js/adapt-contrib-media.js
@@ -11,10 +11,10 @@ define([
   // The following function is used to to prevent a memory leak in Internet Explorer
   // See: http://javascript.crockford.com/memory/leak.html
   function purge(d) {
-    var a = d.attributes, i, l, n;
+    var a = d.attributes;
     if (a) {
-      for (i = a.length - 1; i >= 0; i -= 1) {
-        n = a[i].name;
+      for (let i = a.length - 1; i >= 0; i -= 1) {
+        let n = a[i].name;
         if (typeof d[n] === 'function') {
           d[n] = null;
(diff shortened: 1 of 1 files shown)


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

js/adapt-contrib-media.js
```diff
@@ -38,31 +37,31 @@ define([
    * is a duration integer the other is the player object. The default functions error on slider key press and so break
    * accessibility. Below is a correction.
    */
-  _.extend(mejs.MepDefaults, {
+  _.extend(window.mejs.MepDefaults, {
```
tomgreenfield, line 40: Toying instead with the idea of having a nested `.eslintrc.json` at plugin level i.e.:
```json
{
  "globals": {
    "mejs": true
  }
}
```
Would then nicely match what we do with known libraries at [framework level](https://github.com/adaptlearning/adapt_framework/blob/fb48d28115a357d71cbf3259be9c270cdcff77fa/.eslintrc.json#L11-L22).

moloko: yeah could be a good idea. I still like to include `window.` anyway though.

tomgreenfield (approved):
No refusal from me 👁️

moloko:
thanks @tomgreenfield !

PR #214, opened 2021-03-09
Converted js to es6 classes

eleanor-heath:
Linked issue https://github.com/adaptlearning/adapt_framework/issues/2808


cahirodoherty-learningpool (approved):
👀

js/adapt-contrib-media.js
```diff
@@ -39,34 +37,40 @@ define([
...
-  var MediaView = ComponentView.extend({
+  class MediaView extends ComponentView {
 
-    events: {
-      'click .js-media-inline-transcript-toggle': 'onToggleInlineTranscript',
-      'click .js-media-external-transcript-click': 'onExternalTranscriptClicked',
-      'click .js-skip-to-transcript': 'onSkipToTranscript'
-    },
+    preinitialize() {
+      this.froogaloopAdded = false;
```
oliverfoster, line 53: The variable `froogaloopAdded` is defined in the module because its behaviour should only happen once, only when the first vimeo video is added and not on every component render.

js/adapt-contrib-media.js
```diff
@@ -39,34 +37,40 @@ define([
    */
   _.extend(window.mejs.MepDefaults, {
     keyActions: [],
-    defaultSeekForwardInterval: function(duration) {
+    defaultSeekForwardInterval: (duration) => {
       if (typeof duration === 'object') return duration.duration * 0.05;
       return duration * 0.05;
     },
-    defaultSeekBackwardInterval: function(duration) {
+    defaultSeekBackwardInterval: (duration) => {
```
oliverfoster, line 44:
```suggestion
    defaultSeekBackwardInterval: duration => {
```
Brackets are unnecessary. 

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#comparing_traditional_functions_to_arrow_functions

js/adapt-contrib-media.js
```diff
@@ -173,48 +177,48 @@ define([
...
       if (!media) return callback();
 
       switch (media.type) {
         case 'video/vimeo':
           modelOptions.alwaysShowControls = false;
           modelOptions.hideVideoControlsOnLoad = true;
           modelOptions.features = [];
-          if (froogaloopAdded) return callback();
+          if (this.froogaloopAdded) return callback();
           $.getScript('assets/froogaloop.js')
```
oliverfoster, line 200: You can see here `froogaloop` is imported globally, so this should only be done once globally.

js/adapt-contrib-media.js
```diff
@@ -460,21 +464,21 @@ define([
       }
 
       ComponentView.prototype.remove.call(this);
```
oliverfoster, line 466:
```suggestion
      super.remove();
```

js/adapt-contrib-media.js
```diff
@@ -592,7 +596,7 @@ define([
       Adapt.trigger('media', eventObj);
     }
 
-  });
+  };
 
   return Adapt.register('media', {
```
oliverfoster, line 601: This stuff should be in the main file, and the view should be in a view file.

oliverfoster (changes requested):

eleanor-heath:
@oliverfoster can you take another look when you have a chance please.

js/adapt-contrib-media.js
```diff
@@ -1,598 +1,10 @@
 define([
   'core/js/adapt',
-  'core/js/views/componentView',
+  './mediaView',
   'core/js/models/componentModel',
   'libraries/mediaelement-and-player',
```
oliverfoster, line 5: The libraries are only needed in the view.

js/mediaView.js
```diff
@@ -0,0 +1,598 @@
...
+  /*
+   * Default shortcut keys trap a screen reader user inside the player once in focus. These keys are unnecessary
+   * as one may traverse the player in a linear fashion without needing to know or use shortcut keys. Below is
+   * the removal of the default shortcut keys.
+   *
+   * The default seek interval functions are passed two different data types from mejs which they handle incorrectly. One
+   * is a duration integer the other is the player object. The default functions error on slider key press and so break
+   * accessibility. Below is a correction.
+   */
+  _.extend(window.mejs.MepDefaults, {
```
oliverfoster, line 39: All `_.extend` can become `Object.assign`

js/mediaView.js
```diff
@@ -0,0 +1,598 @@
...
+    events() {
+      return {
+        'click .js-media-inline-transcript-toggle': 'onToggleInlineTranscript',
+        'click .js-media-external-transcript-click': 'onExternalTranscriptClicked',
+        'click .js-skip-to-transcript': 'onSkipToTranscript'
+      };
+    }
+
+    className() {
+      const classes = ComponentView.prototype.className.call(this);
```
oliverfoster, line 62:
```suggestion
      const classes = super.className();
```

js/mediaView.js
```diff
@@ -0,0 +1,598 @@
...
+      if (modelOptions.alwaysShowControls === undefined) {
+        modelOptions.alwaysShowControls = false;
+      }
+      if (modelOptions.hideVideoControlsOnLoad === undefined) {
+        modelOptions.hideVideoControlsOnLoad = true;
+      }
+
+      this.addMediaTypeClass();
+
+      this.addThirdPartyFixes(modelOptions, function createPlayer() {
```
oliverfoster, line 160: Fat arrow

js/mediaView.js
```diff
@@ -0,0 +1,598 @@
...
+        type: eventType,
+        src: this.mediaElement.src,
+        platform: this.mediaElement.pluginType
+      };
+
+      if (player) eventObj.isVideo = player.isVideo;
+
+      Adapt.trigger('media', eventObj);
+    }
+  };
```
oliverfoster, line 595: No semi colon required

oliverfoster (changes requested):

oliverfoster (approved):

moloko:
re: 71ae9c7 you don't actually need the trailing whitespace in all those instances - it's only needed where you need to force a line break - see https://www.markdownguide.org/basic-syntax/#line-breaks

for example, you don't need to have two spaces on a line that's set up as a heading

(no need to change this PR, it's more just FYI for future README amends)

js/mediaView.js
```diff
@@ -0,0 +1,604 @@
...
+    /**
+     * When the learner selects a captions language in another media component, that language may not be available
+     * in this instance, in which case default to the `_startLanguage` if that's set - or "none" if it's not
+     * @param {string} lang The language we're being asked to switch to e.g. "de"
+     * @return {string} The language we're actually going to switch to - or "none" if there's no good match
+     */
+    checkForSupportedCCLanguage(lang) {
+      if (!lang || lang === 'none') return 'none';
+
+      if (_.findWhere(this.model.get('_media').cc, { srclang: lang })) return lang;
```
moloko, line 300: there is an opportunity to switch from `_.findWhere` to vanilla js here... but whether it's really worth it or not is questionable..! I think we should probably leave as-is but if anyone else feels differently, make your case...

js/mediaView.js
```diff
@@ -0,0 +1,604 @@
...
+          modelOptions.features = [];
+          if (froogaloopAdded) return callback();
+          $.getScript('assets/froogaloop.js')
+            .done(() => {
+              froogaloopAdded = true;
+              callback();
+            })
+            .fail(() => {
+              froogaloopAdded = false;
+              console.log('Could not load froogaloop.js');
```
moloko, line 210:
```suggestion
              Adapt.log.error('Could not load froogaloop.js');
```

js/mediaView.js
```diff
@@ -0,0 +1,604 @@
...
+        'media:stop': this.onMediaStop
+      });
+
+      _.bindAll(this, 'onMediaElementPlay', 'onMediaElementPause', 'onMediaElementEnded', 'onMediaElementTimeUpdate', 'onMediaElementSeeking');
+
+      // set initial player state attributes
+      this.model.set({
+        '_isMediaEnded': false,
+        '_isMediaPlaying': false
+      });
```
moloko, line 84:
```suggestion
      this.model.set({
        _isMediaEnded: false,
        _isMediaPlaying: false
      });
```

js/mediaView.js
```diff
@@ -0,0 +1,604 @@
...
+      if (!player) {
+        console.log("Media.setupPlayPauseToggle: OOPS! there's no player reference.");
+        return;
+      }
+
+      // stop the player dealing with this, we'll do it ourselves
+      player.options.clickToPlayPause = false;
+
+      this.onOverlayClick = this.onOverlayClick.bind(this);
+      this.onMediaElementClick = this.onMediaElementClick.bind(this);
```
moloko, line 378: we could just add 'onOverlayClick' and 'onMediaElementClick' to the `_.bindAll` on line 78 then dispense with these two lines

js/mediaView.js
```diff
@@ -0,0 +1,604 @@
...
+
+    onScreenSizeChanged() {
+      this.$('audio, video').width(this.$('.component__widget').width());
+    }
+
+    onSkipToTranscript() {
+      // need slight delay before focussing button to make it work when JAWS is running
+      // see https://github.com/adaptlearning/adapt_framework/issues/2427
+      _.delay(() => {
+        this.$('.media__transcript-btn').a11y_focus();
```
moloko, line 521: Change to new a11y API, I think it should be:
```suggestion
        Adapt.a11y.focus(this.$('.media__transcript-btn'));
```

moloko (changes requested):
a few eslint fixes plus some other suggestions

moloko (approved):
Looks good to me. Gave it a quick test in IE11, FF & Chrome.

moloko:
@oliverfoster do you want to re-review or are you happy with this?

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

Issue #231, opened 2022-04-25
Add role region to encase the controls


oliverfoster:
https://github.com/adaptlearning/adapt-contrib-media/blob/master/js/mediaView.js#L252

joe-replin:
## Bug Description:
The media player container does not convey a meaningful label for screen reader users. Currently, when a screen reader user arrives on the Play button, they do not hear context for what the Play button will control.

## Steps to Reproduce:

    Using a screen reader, open the course prototype.
    Open the standard experience, then locate the Media section
    Focus on the play button for the media player, and note whether any description for the video is provided. 

## Actual Behavior:

See bug description.

## Expected Behavior:

Apply role="region" along with a meaningful label using aria-label to the container for the video. The label should provide a general description for the topic of the video (ex: Introduction Video)

PR #233, opened 2022-05-17
issue/231 - Add region role to video container

joe-replin:
#231 Assigns the media container a role of 'region' instead of 'application'. Also uses the title and displayTitle as a label for aria context when learners focus on the media container.

### DIFF
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -252,7 +252,10 @@ class MediaView extends ComponentView {
     this.$('.media__widget').children('.mejs-offscreen').remove();
     this.$('[role=application]').removeAttr('role tabindex');
     this.$('.mejs-container').attr('role', 'region');
-    this.$('.mejs-container').attr('aria-label', containerLabel);
+    this.$('.mejs-container').attr({
+      role: 'region',
+      'aria-label': containerLabel
+    });
     this.$('[aria-controls]').removeAttr('aria-controls');
     this.$('.mejs-overlay-play').attr('aria-hidden', 'true');
   }
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -248,9 +248,11 @@ class MediaView extends ComponentView {
   }
 
   cleanUpPlayer() {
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster:
Have you tested with Jaws and NVDA @joe-replin ?

js/mediaView.js
```diff
@@ -248,8 +248,11 @@ class MediaView extends ComponentView {
   }
 
   cleanUpPlayer() {
+    const containerLabel = this.model.get('displayTitle') ? this.model.get('displayTitle') : this.model.get('title');
```
oliverfoster, line 251:
```suggestion
    const containerLabel = this.model.get('displayTitle') || this.model.get('title');
```

oliverfoster: I'm not sure if this container label needs to be defined aside from the displayTitle and title as well.

oliverfoster (changes requested):

js/mediaView.js
joe-replin: I was asked to include a meaningful label alongside the role. Something similar was suggested for the MCQ/GMCQ components for reminding learners of the context for the region while focusing: https://github.com/adaptlearning/adapt-contrib-mcq/issues/177

oliverfoster (approved):

js/mediaView.js
```diff
@@ -248,8 +248,14 @@ class MediaView extends ComponentView {
   }
 
   cleanUpPlayer() {
+    const containerLabel = this.model.get('displayTitle') || this.model.get('title');
     this.$('.media__widget').children('.mejs-offscreen').remove();
     this.$('[role=application]').removeAttr('role tabindex');
+    this.$('.mejs-container').attr('role', 'region');
```
eleanor-heath, line 254: dont think this line is needed because the role attribute is set on the element with class mejs-container below

eleanor-heath (approved):

joe-replin:
Done. Thank you, @eleanor-heath!

PR #249, opened 2023-01-26
Fix: contain fullscreen video when in portrait mode (fixes #248)

### LINKED ISSUE #248: Android fullscreen portrait aspect ratio
As a result of https://github.com/adaptlearning/adapt-contrib-media/pull/217, when a video is run fullscreen on an Android device, in portrait orientation, the aspect ratio of the video is incorrect as it is stretched to fill the viewport. This doesn't occur if using native Android controls via `_playerOptions.AndroidUseNativeControls: true`. The issue is restricted to Android only as iOS switches to native controls automatically when run fullscreen.

kirsty-hames:
fixes https://github.com/adaptlearning/adapt-contrib-media/issues/248 

When a video is run fullscreen on an Android device, in portrait orientation, the aspect ratio of the video is incorrect as it is stretched to fill the viewport and distorts the video. 

This doesn't occur if using native Android controls via `_playerOptions.AndroidUseNativeControls: true`. The issue is restricted to Android only as iOS switches to native controls automatically when run fullscreen.

### DIFF
--- a/less/mep-overrides.less
+++ b/less/mep-overrides.less
@@ -72,14 +72,8 @@
 
 // Bug fix for when video aspect ratio is different to viewport
 // --------------------------------------------------
-// as video is set to object-fit: fill, horizontally center the video 
-.mejs-mediaelement {
-  left: 50%;
-  transform: translateX(-50%);
-}
-// allow the height to fill the viewport without distorting the video width
 .orientation-portrait .mejs-container-fullscreen video {
-  width: auto !important;
+  object-fit: contain;
 }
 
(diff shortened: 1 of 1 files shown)


eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

danielghost (approved):
Tested on Android phone and works as expected. Also checked on iPhone, iPad and desktop to ensure there are no impacts. Tested with both native and mediaelement player controls.

PR #263, opened 2023-04-28
Fix: Make 'skip to transcript' button visible on focus (fixes #262)

### LINKED ISSUE #262: Skip transcript button - Not visible when focused
### Subject of the issue
From an accessibility review we had:

>  A "Skip to Transcript" button is available above media players, but does not become visible when focused. For sighted keyboard users who want to refer to the transcript, navigating to the transcript is less efficient as a result.
> 
> Similar to the treatment provided to the skip navigation link, we recommend making the "Skip to Transcript" button visible when focused to ensure users are not confused by the blank focus stop, and to give all keyboard users equivalent ease accessing the transcript.

```
<button class="aria-label js-skip-to-transcript" tabindex="0" aria-label="Skip to transcript"></button>
```

swashbuck:
Fixes #262 

### Fix
* Makes the 'skip to transcript' button visible on keyboard focus.

### DIFF
--- a/less/media.less
+++ b/less/media.less
@@ -11,6 +11,14 @@
     display: none;
   }
 
+  // Skip to transcript button
+  &__skip-to-transcript:not(:focus-visible) {
+    height: 0;
+    padding: 0;
+    margin: 0;
+    overflow: hidden;
+  }
+
   // Cross browser bug fixes
   // --------------------------------------------------
   .me-plugin {
--- a/less/media.less
+++ b/less/media.less
@@ -12,11 +12,18 @@
   }
 
   // Skip to transcript button
-  &__skip-to-transcript:not(:focus-visible) {
-    height: 0;
-    padding: 0;
-    margin: 0;
-    overflow: hidden;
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

joe-allen-89:
If I navigate through a course with the keyboard when getting to the media component the button itself is not visible but is highlighted, when pressing enter the button then flashes visible before jumping to the transcript button. 

Using Windows/Chrome/fw5.30.3/Latest NVDA.

oliverfoster:
> @swashbuck If I navigate through a course with the keyboard when getting to the media component the button itself is not visible but is highlighted, when pressing enter the button then flashes visible before jumping to the transcript button.
> 
> Using Windows/Chrome/fw5.30.3/Latest NVDA.

The button will only every be shown when tabbed to, as it's not visible to click. :focus-visible therefore isn't necessary here, if :focus works, then use that instead.

joe-allen-89 (approved):
That's done it, the button is now visible when tabbing through.

PR #265, opened 2023-06-21
 New: Added event trigger (fixes: #264)

### LINKED ISSUE #264: Volume & Mute Event Trigger
I would like to add event triggers to this plug-in for the purposes of picking up on them for other means to enhance experiences & capabilities. More specifically, I want to track the mute state and volume value of each interaction with the player.

joe-replin:

### New
* Adds trigger to volume controls to indicate the mute state and current volume value
* Fixes #264 

### Testing
1. Install PR and use browser console to test.

### DIFF
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -217,7 +217,7 @@ class MediaView extends ComponentView {
       'media:stop': this.onMediaStop
     });
 
-    _.bindAll(this, 'onMediaElementPlay', 'onMediaElementPause', 'onMediaElementEnded', 'onMediaElementTimeUpdate', 'onMediaElementSeeking', 'onOverlayClick', 'onMediaElementClick', 'onWidgetInview');
+    _.bindAll(this, 'onMediaElementPlay', 'onMediaElementPause', 'onMediaElementEnded', 'onMediaVolumeChange', 'onMediaElementTimeUpdate', 'onMediaElementSeeking', 'onOverlayClick', 'onMediaElementClick', 'onWidgetInview');
 
     // set initial player state attributes
     this.model.set({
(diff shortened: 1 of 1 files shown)


joe-allen-89 (approved):

js/mediaView.js
```diff
@@ -481,6 +482,10 @@ class MediaView extends ComponentView {
     }
   }
 
+  onMediaVolumeChange(event) {
+    Adapt.trigger('media:volumeChange', this, this.mediaElement.muted, this.mediaElement.volume);
```
oliverfoster, line 486: How does this signify which media is changed?
You're passing `[view, muted, volume]`, would `[_id, muted, volume]` not be easier to use where consumed?

joe-replin: Hey @oliverfoster, this is so whatever is on the receiving end has the options available to read more into the model data. Sometimes we'll want available access to the _id, title, lang, displayTitle, model type, etc.
https://github.com/adaptlearning/adapt-contrib-xapi/blob/8b375def92ba6da3eea14e710b734f1a8373c8b4/js/XAPI.js#L670
https://github.com/cgkineo/adapt-xAPI/blob/ea9400ecb5cf0c18487d04dbe5247f849f2c863d/js/statements/abstractStatementModel.js#L24

oliverfoster: Did you mean to pass the model instead of the view?
```suggestion
    Adapt.trigger('media:volumeChange', this.model, this.mediaElement.muted, this.mediaElement.volume);
```

joe-replin: Ah, I see how that would be more efficient. Thanks :)

PR #284, opened 2024-02-08
Update: Adjusts transcript triggers (fixes #283)

### LINKED ISSUE #283: Detecting transcript clicks
Instead of triggering an event when the transcript completes the component, I would additionally like to track transcript toggle interactions.

### Your environment
* v5.35.3

joe-replin:

### Update
* Adds new trigger to capture transcript toggles
* Adjusts transcript completion so it only triggers when it completes the component
* fixes #283

### DIFF
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -506,7 +506,6 @@ class MediaView extends ComponentView {
   }
   
   onExternalTranscriptClicked(event) {
-    if (this.model.get('_transcript')._setCompletionOnView === false) return;
     this.transcriptTriggers();
   }
 
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -518,7 +518,8 @@ class MediaView extends ComponentView {
       return Adapt.trigger('media:transcript', state, this);
     }
     this.setCompletionStatus();
-    Adapt.trigger('media:transcriptComplete', state, this);
+    Adapt.trigger('media:transcript', state, this);
+    Adapt.trigger('media:transcript', 'complete', this);
   }
 
   /**
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -504,16 +504,16 @@ class MediaView extends ComponentView {
     $buttonText.html(this.model.get('_transcript').inlineTranscriptCloseButton);
     this.transcriptTriggers('opened');
   }
-  
+
   onExternalTranscriptClicked() {
     this.transcriptTriggers('external');
   }
-  
+
   transcriptTriggers(state) {
     const setCompletionOnView = this.model.get('_transcript')._setCompletionOnView;
     const isComplete = this.model.get('_isComplete');
     const shouldComplete = (setCompletionOnView && !isComplete);
-    
+
     if (!shouldComplete) {
       return Adapt.trigger('media:transcript', state, this);
     }
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -518,7 +518,6 @@ class MediaView extends ComponentView {
       return Adapt.trigger('media:transcript', state, this);
     }
     this.setCompletionStatus();
-    Adapt.trigger('media:transcript', state, this);
     Adapt.trigger('media:transcript', 'complete', this);
   }
 
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -521,7 +521,7 @@ class MediaView extends ComponentView {
   
     if (setCompletionOnView) {
       const transcriptComplete = isComplete ? 'media:transcript' : 'media:transcriptComplete';
-      Adapt.trigger(transcriptComplete, this);
+      Adapt.trigger(media:transcriptComplete, this);
       this.setCompletionStatus();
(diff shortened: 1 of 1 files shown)


js/mediaView.js
```diff
@@ -501,17 +503,27 @@ class MediaView extends ComponentView {
...
+    const isComplete = this.model.get('_isComplete');
+  
+    if (!setCompletionOnView || (setCompletionOnView && isComplete)) {
+      Adapt.trigger('media:transcript', this);
+      return;
+    } 
+  
+    if (setCompletionOnView) {
+      const transcriptComplete = isComplete ? 'media:transcript' : 'media:transcriptComplete';
+      Adapt.trigger(media:transcriptComplete, this);
```
danielghost, line 524: The event name needs to be a string if being used this way:
```JavaScript
Adapt.trigger(`media:${transcriptComplete}`, this);
```

Not sure the logic makes sense though anyway as it will only ever trigger 'media:transcriptComplete' via this condition. Am assuming this should have just been `Adapt.trigger('media:transcriptComplete', this);`?

js/mediaView.js
```diff
@@ -501,17 +503,27 @@ class MediaView extends ComponentView {
...
+      this.setCompletionStatus();
+    }
```
danielghost, line 526: Can this method be simplified to the following without impact?
```JavaScript
transcriptTriggers() {
  const setCompletionOnView = this.model.get('_transcript')._setCompletionOnView;
  const isComplete = this.model.get('_isComplete');
  Adapt.trigger('media:transcript', this);
  if (setCompletionOnView && !isComplete) {
    Adapt.trigger('media:transcriptComplete', this);
    this.setCompletionStatus();
  }
}
```

js/mediaView.js
```diff
@@ -501,17 +503,27 @@ class MediaView extends ComponentView {
...
-
-    if (this.model.get('_transcript')._setCompletionOnView !== false) {
-      Adapt.trigger('media:transcriptComplete', this);
-      this.setCompletionStatus();
-    }
   }
-
+  
   onExternalTranscriptClicked(event) {
     if (this.model.get('_transcript')._setCompletionOnView === false) return;
```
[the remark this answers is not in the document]
danielghost: Yes this need to be removed otherwise it will only trigger if `_setCompletionOnView: true`.

danielghost (changes requested):

js/mediaView.js
oliverfoster: You can reduce nesting and increase readability by doing this:
```JavaScript
transcriptTriggers() {
  const setCompletionOnView = this.model.get('_transcript')._setCompletionOnView;
  const isComplete = this.model.get('_isComplete');
  const shouldComplete = (setCompletionOnView && !isComplete);
  Adapt.trigger('media:transcript', this);
  if (!shouldComplete) return;
  this.setCompletionStatus();
  Adapt.trigger('media:transcriptComplete', this);
}
```

js/mediaView.js
```diff
@@ -485,6 +485,8 @@ class MediaView extends ComponentView {
     const $button = this.$('.media__transcript-btn-inline');
     const $buttonText = this.$('.media__transcript-btn-inline .media__transcript-btn-text');
 
+    this.transcriptTriggers();
```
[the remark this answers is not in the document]
joe-replin: I weighed both sides together and leaned this way.ith other media UI such as volume or speed. At a high-level, these triggers are collecting click rates.
1. If tracking only the on open toggle, it's probably better to restrict that in an alternate plug-in based on your needs.
2. I was also keeping inline with how the plugin simply triggers clicks in other media UI such as volume & speed.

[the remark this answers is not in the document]
danielghost: In that case, can it be amended to include the transcript state with the trigger?

danielghost (changes requested):

js/mediaView.js
```diff
@@ -503,22 +502,23 @@ class MediaView extends ComponentView {
...
     const isComplete = this.model.get('_isComplete');
     const shouldComplete = (setCompletionOnView && !isComplete);
     
     if (!shouldComplete) {
-      return Adapt.trigger('media:transcript', this);
+      return Adapt.trigger('media:transcript', state, this);
     }
     this.setCompletionStatus();
-    Adapt.trigger('media:transcriptComplete', this);
+    Adapt.trigger('media:transcriptComplete', state, this);
```
oliverfoster, line 521:
```suggestion
    Adapt.trigger('media:transcript', state, this);
    Adapt.trigger('media:transcript', 'complete', this);
```

There seems no point having a separate event now that there's a state component.

oliverfoster (approved):

swashbuck (approved):
👍

oliverfoster:
@cahirodoherty-learningpool would you mind rechecking this please?

cahirodoherty-learningpool (approved):
👀

Issue #286, opened 2024-02-28
End to end tests are needed for the media component


lemmyadams:
### Subject of the issue/enhancement/features
Now that e2e tests are run on PR creation we should add to these. This issue is specifically directed at tests for the media component

### Your environment
* >= [v5.30.0](https://github.com/adaptlearning/adapt_framework/releases/tag/v5.30.0)

PR #287, opened 2024-02-28
Chore: e2e tests on media component (issue/286)

lemmyadams:
Addresses #286

### DIFF
--- a/test/e2e/media.cy.js
+++ b/test/e2e/media.cy.js
@@ -1,0 +1,34 @@
+describe('Media', function () {
+  beforeEach(function () {
+    cy.getData()
+  });
+
+  it('should display the media component', function () {
+    const mediaComponents = this.data.components.filter((component) => component._component === 'media')
+    mediaComponents.forEach((mediaComponent) => {
+      cy.visit(`/#/preview/${mediaComponent._id}`);
+      const bodyWithoutHtml = mediaComponent.body.replace(/<[^>]*>/g, '');
+      
+      cy.testContainsOrNotExists('.media__title', mediaComponent.displayTitle)
+      cy.testContainsOrNotExists('.media__body', bodyWithoutHtml)
+      cy.testContainsOrNotExists('.media__instruction', mediaComponent.instruction)
+
+      if(mediaComponent._media.mp4) {
+        cy.get('.mejs-mediaelement video').should('have.attr', 'src', mediaComponent._media.mp4)
+      }
+      if(mediaComponent._media.poster) {
+        cy.get('.mejs-poster img').should('have.attr', 'src', mediaComponent._media.poster)
+      }
+
+      if(mediaComponent._transcript) {
+        cy.get('.media__transcript-body-inline').should('not.be.visible')
+        cy.get('button.media__transcript-btn').should('contain', mediaComponent._transcript.inlineTranscriptButton).click()
+        cy.get('.media__transcript-body-inline-inner').should('be.visible').should('contain', mediaComponent._transcript.inlineTranscriptBody)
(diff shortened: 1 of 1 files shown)


test/e2e/media.cy.js
```diff
@@ -0,0 +1,34 @@
...
+
+  it('should display the media component', function () {
+    const mediaComponents = this.data.components.filter((component) => component._component === 'media')
+    mediaComponents.forEach((mediaComponent) => {
+      cy.visit(`/#/preview/${mediaComponent._id}`);
+      const bodyWithoutHtml = mediaComponent.body.replace(/<[^>]*>/g, '');
+      
+      cy.testContainsOrNotExists('.media__title', mediaComponent.displayTitle)
+      cy.testContainsOrNotExists('.media__body', bodyWithoutHtml)
+      cy.testContainsOrNotExists('.media__instruction', mediaComponent.instruction)
```
eleanor-heath, line 14: Minor: theres quite a few missing semicolons in the code.

test/e2e/media.cy.js
```diff
@@ -0,0 +1,34 @@
...
+
+      if(mediaComponent._media.mp4) {
```
eleanor-heath, line 16: Missing space beside if, same in the if statements below.

test/e2e/media.cy.js
```diff
@@ -0,0 +1,34 @@
+describe('Media', function () {
+  beforeEach(function () {
+    cy.getData()
+  });
+
+  it('should display the media component', function () {
```
eleanor-heath, line 6: could use arrow function here instead of function

test/e2e/media.cy.js
```diff
@@ -0,0 +1,34 @@
...
+
+      if(mediaComponent._transcript) {
+        cy.get('.media__transcript-body-inline').should('not.be.visible')
+        cy.get('button.media__transcript-btn').should('contain', mediaComponent._transcript.inlineTranscriptButton).click()
+        cy.get('.media__transcript-body-inline-inner').should('be.visible').should('contain', mediaComponent._transcript.inlineTranscriptBody)
+        cy.get('button.media__transcript-btn').should('contain', mediaComponent._transcript.inlineTranscriptCloseButton).click()
+        cy.get('.media__transcript-body-inline').should('not.be.visible')
+      }
+
+      cy.wait(1000)
```
eleanor-heath, line 31: if we are waiting for a XHR request to complete here it might be worth adding in a new cypress command that does just that rather than waiting a random amount of time, e.g.

```
Cypress.Commands.add(('waitForXHR'), (path, varName) => {
  cy.intercept(path).as(varName);
  cy.wait(`@${varName}`, { timeout: 30000 });
});
```

eleanor-heath (approved):

test/e2e/media.cy.js
oliverfoster: these should be being linted with eslint in the same way as the rest of adapt's code

test/e2e/media.cy.js
oliverfoster: You need to leave these as `function` for the `this` context which includes `this.data` I believe. A weird cypress thing.

test/e2e/media.cy.js
oliverfoster: Pr to fix https://github.com/adaptlearning/adapt_framework/pull/3522

test/e2e/media.cy.js
oliverfoster: I think we're waiting for the component to be rendered in a new adapt page. Which is much harder to detect from cypress.

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #291, opened 2024-04-11
New: completion icon added to transcript button (fixes #290)

### LINKED ISSUE #290: Media transcript completion styling relies on colour alone
### Subject of the enhancement
In a previous [Vanilla PR](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/475), selected and visited (completed) styling was added to the media transcript button that mirrors the Accordion item styling. This information will be difficult or impossible to perceive for individuals with visual disabilities such as low vision or color blindness.

### Expected behaviour
Provide a non-color dependent means of identifying completion. 

Display a checkmark icon as per Accordion and Hotgraphic items.

accordion

hotgraphic

Suggestion:

media

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-media/issues/290

### New
* On completion, a checkmark icon is displayed on the transcript button

media

### Update
* Transcript button styled for consistency with `.nav__btn` text label/icon
* LESS styles structured in DOM order (skip to transcript btn, transcript btn, transcript body)

**Not included in PR**
Do you think we should display a plus icon for the inline transcript button to convey the button toggles/expands to reveal content? This would be consistent with the Accordion item button however I'm not wedded to this if people think it's unnecessary.  See example below.

inline_transcript

### DIFF
--- a/less/media.less
+++ b/less/media.less
@@ -7,10 +7,6 @@
   
   // Transcript
   // --------------------------------------------------
-  &__transcript-body-inline {
-    display: none;
-  }
-
   // Skip to transcript button
   &__skip-to-transcript {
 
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👍 

> **Not included in PR**
> Do you think we should display a plus icon for the inline transcript button to convey the button toggles/expands to reveal content? This would be consistent with the Accordion item button however I'm not wedded to this if people think it's unnecessary.

I think it would make sense for a plus icon to be present especially as we have similar behaviour in the accordion.

kirsty-hames:
> Yes to + buttons for expand/collapse generally, maybe not for external transcript? Perhaps? Dunno.

Agreed. The external transcript button doesn't expand so there is no need for a + icon. New issue raised to display plus icon for inline transcript button - https://github.com/adaptlearning/adapt-contrib-media/issues/292.

Issue #297, opened 2024-07-17
Video Speed menu is not accessible


swashbuck:
### Subject of the issue
The `_speed` player option is not accessible:

1. The speed selector button does not have a meaningful label for screenreaders. The button is announced by default as the "one point zero zero x button"
2. The speed selector button is not keyboard accessible. Mouse users can hover over the playback speed button to reveal more options, but when a keyboard users sets focus on the playback speed button and activates it with space or enter, nothing happens. Keyboard users are unable to change their playback speed.

In _mediaLibrariesOverrides.js_, we [override](https://github.com/adaptlearning/adapt-contrib-media/blob/a9f2bc7c252e772390b2533a1b7d228eb32e32c5/js/mediaLibrariesOverrides.js#L265) (or add?) aria labels for use with MediaElement.js. So, perhaps we could add a label for speed here.

The focus issue is similar to that of captions. See #232 

### Steps to reproduce
Add 'speed' to a Media component's player options. For example:

```
"_playerOptions": {
  "alwaysShowControls": false,
  "hideVideoControlsOnLoad": true,
  "showPosterWhenEnded": true,
  "features": [
    "playpause",
    "progress",
    "fullscreen",
    "volume",
    "speed"
  ]
},
```

Then, navigate to the option with a screenreader / keyboard navigation.

swashbuck:
Created #298 to look into updating the MediaElement library.

swashbuck:
Needs reviewing as part of #299

kirsty-hames:
> Needs reviewing as part of [#299](https://github.com/adaptlearning/adapt-contrib-media/pull/299)

Similar to the [caption button issue](https://github.com/adaptlearning/adapt-contrib-media/issues/232#issuecomment-3601905875) the video speed list (menu for video speed selection) doesn't display when focused. The list only displays on hover.

This is due to the `.mejs__offscreen` class appended to `.mejs__speed-selector`.

Issue #232, opened 2022-05-13
Cannot navigate caption button language menu with accessibility 


robertmausser:
Caption button is not accessible through tab/ arrow keys. When user hits enter on it then nothing happens.

Caption button should be accessible through tab/ arrow keys. When user hits enter on it then the button should expand.

If caption button will not be accessible then screen reader and keyboard user will not be able to access its functionality.

https://www.screencast.com/t/mLvG21vtC

Provide proper tab indexing and aria role to caption button.

swashbuck:
We can review this again as part of #299

swashbuck:
Needs reviewing as part of #299

kirsty-hames:
> Needs reviewing as part of [#299](https://github.com/adaptlearning/adapt-contrib-media/pull/299)

I've retested this since the PR and captions are still not fully accessible. I can toggle captions on/off fine using keyboard controls but the caption selector list (menu for caption languages) doesn't display when focused. Whilst I can tab onto the list and use arrow keys to select a language, the list is visually hidden but occasionally flickers display when navigating through the options. See screen recording below.

https://github.com/user-attachments/assets/11e785c4-2f3b-4ae3-810c-2beb197356cd

This is due to the `.mejs__offscreen` class appended to `.mejs__captions-selector`. [Issue has been raised for mediaelement](https://github.com/mediaelement/mediaelement/issues/2949). 

The caption selector list should remain on screen whilst it has focus. We could probably patch fix this with a CSS override until it's addressed in mediaelement.

swashbuck:
> We could probably patch fix this with a CSS override until it's addressed in mediaelement.

@kirsty-hames I think that's a good solution considering the infrequency of mediaelement releases. We can add a comment regarding the existing [issue](https://github.com/mediaelement/mediaelement/issues/2949) in the Less.

PR #299, opened 2024-08-14
Breaking: Upgrade MediaElement.js to 7.0.7 (fixes #298 #297 #232)

### LINKED ISSUE #298: MediaElement.js upgrade analysis
### Subject of the enhancement
Explore updating the MediaElement.js library. The primary goal is to fix some existing a11y issues like #297 and #232. It is unclear if updating the library will fix those. New features should not be considered as part of this analysis.

### Your environment
* Media 6.7.0

swashbuck:
Fix #298 #297 #232

### Dependencies
* **Requires** Vanilla theme updates from https://github.com/adaptlearning/adapt-contrib-vanilla/pull/524. The exact Vanilla version is TBD, but we should probably add it to the readme.
* **Requires** Framework version [5.42.7](https://github.com/adaptlearning/adapt_framework/releases/tag/v5.42.7) / Core [6.60.8](https://github.com/adaptlearning/adapt-contrib-core/releases/tag/v6.60.8) for usage of `.visually-hidden-focusable` #324

### Breaking
* Upgrades MediaElement.js from v2.21.2 to [v7.0.7](https://github.com/mediaelement/mediaelement/releases/tag/7.0.7). This includes breaking [namespace changes](https://github.com/mediaelement/mediaelement/blob/master/MIGRATION.md#migrating-from-2x-to-4x-version).
* Removes Flash [support](https://github.com/mediaelement/mediaelement/blob/master/MIGRATION.md#migrating-from-6x-to-7x-version)
* Removes IE 9/10 [support](https://github.com/mediaelement/mediaelement/blob/master/MIGRATION.md#changes-on-420-version)
* Removes Vimeo support (Vimeo-specific styles and JS shims). Anyone who needs Vimeo support should migrate to the Adapt [Vimeo plugin](https://github.com/adaptlearning/adapt-vimeo).
* Removes YouTube support. Anyone who needs YouTube support should migrate to the Adapt [YouTube plugin](https://github.com/adaptlearning/adapt-youtube).

### Notable changes and deletions
* First off, refer to the MediaElement [migration notes](https://github.com/mediaelement/mediaelement/blob/master/MIGRATION.md#migrating-from-2x-to-4x-version), particularly regarding the [namespace changes](https://github.com/mediaelement/mediaelement/blob/master/MIGRATION.md#migrating-from-2x-to-4x-version).
* `speed`, `jumpforward` and `skipback` have moved to mejs plugins. Other [plugins](https://github.com/mediaelement/mediaelement-plugins/tree/master?tab=readme-ov-file#available-plugins) can be considered for inclusion in the future.
* The `MediaFeatures` object is now just `Features`
* MediaElement success callback passes different parameters: `media`, `node`, and `instance`. I've added documentation about this to `onPlayerReady()` which is the success callback function.
* Some properties have moved. `mediaElement.player` changed to `mediaElement.options`. `selectedTrack` has moved to the `instance` object mentioned above.
* Removed many assets. Icon implementation by default now uses SVG [symbols](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/symbol). When using the Vanilla theme (PR listed below), these icons are replaced with Vanilla font icons.
* Added a README.md to the libraries directory detailing the changes we've directly made to the mejs library itself. These changes have pending PRs at the mejs GitHub repo.

### Fix
* #297 
* #232 

### Configuration changes to existing courses
None

### Regression testing
The following fixes could use retesting. I could not find the original GitHub tickets for some of these.

1. "**Default shortcut keys trap a screen reader user** inside the player once in focus. These keys are unnecessary as one may traverse the player in a linear fashion without needing to know or use shortcut keys. Below is the removal of the default shortcut keys." The fix I've implemented is slightly different from the [old fix](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/js/mediaLibrariesOverrides.js#L3).
2. Removed: "The **default seek interval functions** are passed two different data types from mejs which they handle incorrectly. One is a duration integer the other is the player object. The default functions error on slider key press and so break accessibility." The [old fix](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/js/mediaLibrariesOverrides.js#L3) rewrites the defaultSeekForwardInterval and defaultSeekBackwardInterval  functions to check the type of the input. These two functions have been slightly rewritten in the newer mejs library, so a fix may no longer be needed. I have removed the previous fix for now.
3. Removed [**setupPlayPauseToggle()**](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/js/mediaView.js#L333) fix: "Overrides the default play/pause functionality to stop accidental playing on touch devices... we don't have a this.mediaElement.player ref on iOS devices." The fix seemed to actually break play/pause functionality on iOS, so I've removed it.
4. Removed [**buildfullscreen()**](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/js/mediaLibrariesOverrides.js#L60) override ("Overwrite mediaelement-and-player buildfullscreen to remove global delete"). The only significant difference in the current override was to remove two `delete hideTimeout` lines. The new `buildfullscreen()` in `mediaelement-and-player.js` does not use this.
5. Removed [**killContextMenuTimer**](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/js/mediaLibrariesOverrides.js#L46) override to "remove global delete" since it no longer exists in mejs
6. Removed CSS [fix]() for https://github.com/adaptlearning/adapt_framework/issues/1883
7. Removed CSS [fix](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/less/mep-overrides.less#L16) for "touch devices to override standard hover event in mep.less L119"
8. Removed CSS [fix](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/less/mep-overrides.less#L79C4-L79C58) for "missing control icons on iOS touch devices"
9. Removed CSS [fix](https://github.com/adaptlearning/adapt-contrib-media/blob/7fb7d7ef6ee5762e130f7e672baf1bff7c45704c/less/mep-overrides.less#L4C6-L5C87) for "The width set by mejs is always 100% of container, set in pixels. Having a fixed pixel size prevents iOS from changing orientation without scaling."
10. Fix for firefox fullscreen api #239 The fix for this PR was recent, so it should still be fine.

### Known issues
* #304


simondate:
I've noticed that Youtube isn't working either - The README does mention support is temperamental, but perhaps this should be properly documented as being removed like it is with Vimeo.

swashbuck:
> I've noticed that Youtube isn't working either - The README does mention support is temperamental, but perhaps this should be properly documented as being removed like it is with Vimeo.

Thanks, @simondate . I've added this to the PR notes above.

swashbuck:
Ready for review.

lemmyadams (approved):

oliverfoster:
On any existing course, when updating adapt-contrib-media, **without** also including the vanilla pull request [#524](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/524), the media element will not be styled, it will look as the default mejs styling:
image

Once the file changes from the vanilla pull request are included, mep.less and mep-legacy.less, it should look like this:
image

New courses with this and the vanilla pr will be styled as above and expected.

oliverfoster:
Further bug fixes: https://github.com/adaptlearning/adapt-contrib-media/pull/311

oliverfoster:
None of these files need executable permissions because none of them are executable. What are the last commits trying to fix?

swashbuck:
> None of these files need executable permissions because none of them are executable. What are the last commits trying to fix?

I was reverting the change made in this PR to get the file permissions back to where they were previously. For example, [adapt-contrib-media.js](https://github.com/adaptlearning/adapt-contrib-media/blob/master/js/adapt-contrib-media.js) is currently executable. I didn't want to add any additional files to this already overstuffed PR. Happy to change it, though?

Screenshot 2025-03-20 at 8 18 06 AM

swashbuck:
**Bug:** 🐞 Need to fix an issue in Chrome. The closed captions are not showing when in fullscreen mode. Firefox seems fine. Testing on a Mac.

swashbuck:
> **Bug:** 🐞 Need to fix an issue in Chrome. The closed captions are not showing when in fullscreen mode. Firefox seems fine. Testing on a Mac. Need to see if this is a bug with the ME.js library.

Fixed by https://github.com/adaptlearning/adapt-contrib-vanilla/pull/524/commits/2188c7d3691f9702e23bc803d777395bcb468f10

oliverfoster:
Is this ready for review? @swashbuck

swashbuck:
> Is this ready for review? @swashbuck

It is, aside from any new issues we find.

js/mediaView.js
```diff
@@ -59,26 +73,36 @@ class MediaView extends ComponentView {
...
-      });
-    }
+    this.addThirdPartyFixes(modelOptions, () => {
+      // create the player
+      this.$('audio, video').mediaelementplayer(modelOptions);
+      this.cleanUpPlayer();
+
+      const _media = this.model.get('_media');
+      // if no media is selected - set ready now, as success won't be called
+      if (!_media.mp3 && !_media.mp4 && !_media.ogv && !_media.webm && !_media.source) {
```
oliverfoster, line 91: Probably need similar if someone tries to use youtube or vimeo.

templates/media.hbs
```diff
@@ -22,10 +22,8 @@
     {{else}}
       <video aria-hidden="true" preload="none"
         {{#equals _aspectRatio "square"}} width="640" height="640"
-        {{else equals _aspectRatio "portrait"}} width="540" height="960"
-        {{else}} width="640" height="360"
-        {{/equals}}
-        {{#if _playsinline}} playsinline{{/if}} {{#equals _media.type "video/vimeo"}}{{else}}poster="{{_media.poster}}"{{/equals}} style="width:100%; height:100%;" controls="controls">
+        {{else equals _aspectRatio "portrait"}} width="540" height="960"{{/equals}}
```
oliverfoster, line 25: Was there a reason for removing the default landscape sizing?

oliverfoster (changes requested):
We should probably remove `transcript.pdf` from the `assets/` folder to prevent it from being unnecessarily distributed with every course containing this plugin. It can be moved to `adapt_framework`.

joe-allen-89 (commented):
Testing using NVDA and latest Firefox/Chrome, couple of issues, but looking really good overall
- You can navigate through the 'mejs__captions-selector' when it's not visible
- On chrome there's a weird bug that if you open captions and select none, you can't use the arrows to navigate away
- When navigating through the selector it highlights an invisible checkbox with 'radio button not checked', then jumps to the text which doesn't include if it's selected
- Skip to transcript button missing in firefox (works as expected in Chrome)
- When exiting fullscreen, the screenreader jumps to the top of the page (this was also an issue with latest media release)

swashbuck:
> 🪲 **New bug** Found an issue where backwards scrubbing is broken #323 Should fix _or_ revert this before release.

I've reverted part of the #320 fix. Backwards scrubbing is now possible when preventing forward scrubbing.

libraries/README.md
```diff
@@ -0,0 +1,8 @@
+# Modifications
+
+## mediaelement-and-player.js
```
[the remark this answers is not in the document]
oliverfoster: Could you do a permalink to a version or commit or tag instead of the current matter branch?

oliverfoster:
@joe-allen-89 

> When exiting fullscreen, the screenreader jumps to the top of the page (this was also an issue with latest media release)

@danielghost has a core fix for this hopefully

libraries/README.md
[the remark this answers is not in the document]
swashbuck: Done in [2cc67fd](https://github.com/adaptlearning/adapt-contrib-media/pull/299/commits/2cc67fd56c69f29276b1e46b3bc30c36908dd013)

templates/media.hbs
swashbuck: Ok, I've set this up to _not_ set a size when using native controls. Could use some help with Android testing, though.

oliverfoster: What do you need?

swashbuck: Testing on an Android device with `AndroidUseNativeControls` set to true and false

swashbuck:
> Skip to transcript button missing in firefox (works as expected in Chrome)

@joe-allen-89 Will you retest this? Should be fixed now.

templates/media.hbs
oliverfoster: Both the players are working fine on Pixel 7, Chrome 138.0.7151.72 Android 15

oliverfoster:
image

Firefox skip to transcript working here.

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

PR #320, opened 2025-03-22
Fix: Preventing forward scrubbing (fixes: #319)

### LINKED ISSUE #319: Fully preventing forward scrubbing
It seems that learners are bypassing the `_preventForwardScrubbing` setting by interacting with the locked media player scrub bar. Despite the setting being enabled, a small number of learners repeatedly select the scrub bar and are able to skip forward in the video, essentially bypassing the need to watch the content in full.

The issue likely stems from the current solution, which forces the scrub bar to the furthest point of progress, but this approach isn't fully effective. A more robust solution would be to ensure that the locked scrub bar is not interactable or selectable at all.

### Your environment
* FW v5.35.3
* Media v6.6.0

### Steps to reproduce
I'm not entirely sure how to reliably reproduce this issue. It seems to occur intermittently with a small subset of learners, but they are able to repeatedly bypass the locked media player by interacting with the scrub bar.

### Expected behaviour
When `_preventForwardScrubbing` is enabled, the scrub bar should be completely disabled, preventing any interaction, including clicks, dragging, or selection. The scrub bar should be non-interactive for all learners.

### Actual behaviour
Currently, the scrub bar remains selectable by mouse, allowing users to interact with it and potentially bypass media content, even with `_preventForwardScrubbing` enabled.

joe-replin:

It seems that learners are bypassing the `_preventForwardScrubbing` setting by interacting with the locked media player scrub bar. Despite the setting being enabled, a small number of learners repeatedly select the scrub bar and are able to skip forward in the video, essentially bypassing the need to watch the content in full.

The issue likely stems from the current solution, which forces the scrub bar to the furthest point of progress, but this approach isn't fully effective. A more robust solution would be to ensure that the locked scrub bar is not interactable or selectable at all.

### Fix
* Fixes #319 
* No pointer events prevents mouse and touch events.
* Prevent default blocks the scrub bar from being changed programmatically.
* aria-disabled makes the scrub bar non-focusable.

I'm only taking a good stab at this being a good solution to truly lock down forward scrubbing. As stated in #319, I'm not sure how the current locks are being bypassed. 

### Testing
Try to break it!

### DIFF
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -192,20 +192,7 @@ class MediaView extends ComponentView {
       this.setupInviewCompletion('.component__widget');
     }
 
-    // wrapper to check if preventForwardScrubbing is turned on.
-    if ((this.model.get('_preventForwardScrubbing')) && (!this.model.get('_isComplete'))) {
-      $(this.mediaElement).on({
-        seeking: this.onMediaElementSeeking,
-        timeupdate: this.onMediaElementTimeUpdate
-      });
-
-      const timeSlider = this.$('.mejs__time-slider')[0];
-      if (timeSlider) {
-        timeSlider.style.pointerEvents = 'none';
-        timeSlider.addEventListener('input', (e) => e.preventDefault());
-        timeSlider.setAttribute('aria-disabled', 'true');
-      }
-    }
+    this.preventForwardScrubbing();
 
     // handle other completion events in the event Listeners
     $(this.mediaElement).on({
@@ -526,6 +513,25 @@ class MediaView extends ComponentView {
     }
     this.setCompletionStatus();
     Adapt.trigger('media:transcript', 'complete', this);
+  }
+
+  /**
+   * This function ensures that users cannot skip ahead in the media until they have watched it fully.
+   */
+  preventForwardScrubbing() {
+    if ((this.model.get('_preventForwardScrubbing')) && (!this.model.get('_isComplete'))) {
+      $(this.mediaElement).on({
+        seeking: this.onMediaElementSeeking,
+        timeupdate: this.onMediaElementTimeUpdate
+      });
+
+      const timeSlider = this.$('.mejs__time-slider')[0];
+      if (timeSlider) {
+        timeSlider.style.pointerEvents = 'none';
+        timeSlider.addEventListener('input', (e) => e.preventDefault());
+        timeSlider.setAttribute('aria-disabled', 'true');
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster (commented):
Would you not want to use all methods, just for peace of mind? And seeing as the code is there still.

joe-replin:
Added back in!

joe-replin:
@swashbuck, please see latest changes for readability and review.

js/mediaView.js
```diff
@@ -521,6 +515,25 @@ class MediaView extends ComponentView {
     Adapt.trigger('media:transcript', 'complete', this);
   }
 
+  /**
+   * This function ensures that users cannot skip ahead in the media until they have watched it fully if `_preventForwardScrubbing` is enabled.
+   */
+  preventForwardScrubbing() {
+    if ((this.model.get('_preventForwardScrubbing')) && (!this.model.get('_isComplete'))) {
```
oliverfoster, line 522: Return early

js/mediaView.js
```diff
@@ -521,6 +515,25 @@ class MediaView extends ComponentView {
...
+        timeupdate: this.onMediaElementTimeUpdate
+      });
+
+      const timeSlider = this.$('.mejs__time-slider')[0];
+      if (timeSlider) {
+        timeSlider.style.pointerEvents = 'none';
+        timeSlider.addEventListener('input', (e) => e.preventDefault());
+        timeSlider.setAttribute('aria-disabled', 'true');
+      }
+    }
```
oliverfoster, line 534: return early:
```suggestion
    if (!this.model.get('_preventForwardScrubbing') || this.model.get('_isComplete')) return;
    $(this.mediaElement).on({
      seeking: this.onMediaElementSeeking,
      timeupdate: this.onMediaElementTimeUpdate
    });
    const timeSlider = this.$('.mejs__time-slider')[0];
    if (timeSlider) {
      timeSlider.style.pointerEvents = 'none';
      timeSlider.addEventListener('input', (e) => e.preventDefault());
      timeSlider.setAttribute('aria-disabled', 'true');
    }
```

oliverfoster (approved):

js/mediaView.js
```diff
@@ -521,6 +515,23 @@ class MediaView extends ComponentView {
...
+    $(this.mediaElement).on({
+      seeking: this.onMediaElementSeeking,
+      timeupdate: this.onMediaElementTimeUpdate
+    });
+    const timeSlider = this.$('.mejs__time-slider')[0];
+    if (timeSlider) {
+      timeSlider.style.pointerEvents = 'none';
+      timeSlider.addEventListener('input', (e) => e.preventDefault());
+      timeSlider.setAttribute('aria-disabled', 'true');
+    }
```
swashbuck, line 532: Return early here
```suggestion
    if (!timeSlider) return;
    timeSlider.style.pointerEvents = 'none';
    timeSlider.addEventListener('input', (e) => e.preventDefault());
    timeSlider.setAttribute('aria-disabled', 'true');
```

swashbuck (approved):
👍

PR #329, opened 2025-07-23
Fix: Switch 'WebVTT source' AAT asset filter to 'other' (fixes #328)

### LINKED ISSUE #328: Update CC and transcript asset type filter
Currently CC and transcript files are limited to 'text' mimetypes, which restricts various file types from being selectable in the AAT - such as PDF (transcript), SBV and SRT (subtitles). 

Simplest solution to this is to revert to using 'other', which will display anything not specified as an audio/image/video file.

taylortom:

Fixes #328

### Fix
* Allows all non audio/image/video files to be selectable in the AAT for the 'WebVTT source' field

### DIFF
--- a/schema/component.schema.json
+++ b/schema/component.schema.json
@@ -268,7 +268,7 @@
               "title": "Linked transcript file source",
               "_backboneForms": {
                 "type": "Asset",
-                "media": "text"
+                "media": "other"
               }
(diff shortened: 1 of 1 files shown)


taylortom:
Original issue closed as wontfix.

PR #331, opened 2025-07-23
Fix: Switch transcript AAT asset filter to 'other' (fixes #330)

### LINKED ISSUE #330: Switch transcript AAT asset filter to 'other'
Currently transcript files are limited to 'text' mimetypes, which restricts various file types from being selectable in the AAT - PDF being a common example.

Simplest solution to this is to revert to using 'other', which will display anything _not_ specified as an audio/image/video file.

taylortom:

Fixes #330

### Fix
* Allows all non audio/image/video files to be selectable in the AAT for the transcript files

### DIFF
--- a/schema/component.schema.json
+++ b/schema/component.schema.json
@@ -268,7 +268,7 @@
               "title": "Linked transcript file source",
               "_backboneForms": {
                 "type": "Asset",
-                "media": "text"
+                "media": "other"
               }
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #334, opened 2025-07-29
New: Add forward scrubbing prevention (fix #323)

### LINKED ISSUE #323: Backwards scrubbing is broken
### Subject of the issue
Backwards scrubbing is broken on #299 . Caused by #320 which was meant to prevent forward scrubbing. Only affects PR #299 due to using the newer `mejs__` prefix.

https://github.com/adaptlearning/adapt-contrib-media/blob/89a57baa8a8ca62e80225eade4c28b741c9e8be9/js/mediaView.js#L527-L531

### Your environment
* Media PR #299 

### Steps to reproduce
Click on the scrub bar to go to an earlier point of the video.

### Expected behaviour
It should rewind and play.

### Actual behaviour
It does not.

### LINKED ISSUE #333: Locked player bar still triggers events
A media element with _preventForwardScrubbing still fires events which may lead to some media analytics bloat.

### Your environment
* Media v7.0.0

### Expected behaviour
When _preventForwardScrubbing is enabled, the seeking event should not trigger

### Actual behaviour
The media element fires events in the following order: 
* Pause
* Play
* Seek

joe-replin:

https://github.com/adaptlearning/adapt-contrib-media/pull/320 - was not a comprehensive solution for preventing forward scrubbing. Making another attempt.

### Fix
- Fixes #323 
- Fixes #333
- No pointer events prevents mouse and touch events.
- Prevent default blocks the scrub bar from being changed programmatically.
- aria-disabled makes the scrub bar non-focusable.
- Prevents events from firing when attempting to scrub passed a locked segment of the video.

### Update
- Adjusted `_preventForwardScrubbing` configuration option that prevents learners from skipping ahead in audio/video content until they have viewed/listened to it completely at least once
- Learners can still skip backwards to any previously viewed portion of the media
- Visual blocker overlay appears on the progress bar to indicate the restricted area. Animation provides feedback when learners attempt to scrub forward beyond their max viewed position
- Keyboard navigation (ArrowRight, End, PageDown) is also restricted when at the maximum viewed position
- Restriction is automatically lifted once the component is marked as complete
- Event listener management for forward scrubbing prevention to ensure proper cleanup on component removal
- README, Schemas & Example

### New

### Testing
1. Enable `_preventForwardScrubbing: true` in a media component
2. Set `_setCompletionOn: "ended"`
3. Play the media and verify:
  - You cannot click or drag the progress bar ahead of the current max viewed position
  - You can click/drag backwards to any previously viewed position
  - Arrow keys (ArrowRight, End, PageDown) do not skip ahead when at max viewed position
  - A visual blocker appears on the progress bar showing the restricted area
  - Flash animation displays when attempting to scrub forward
  - After reaching the end and completing the component, all restrictions are removed
4. Test with both audio and video media
5. Verify proper cleanup by navigating away from the page and back

### DIFF
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -547,7 +547,6 @@ class MediaView extends ComponentView {
     // Create and setup the scrub blocker
     const scrubBlocker = this.createScrubBlocker($slider[0]);
 
-    // Setup event handlers
     this.setupScrubBlockerEvents(player, scrubBlocker, () => maxViewed, (newMax) => {
       maxViewed = newMax;
       this.model.set('_maxViewed', maxViewed);
@@ -563,7 +562,6 @@ class MediaView extends ComponentView {
     const scrubBlocker = document.createElement('div');
     scrubBlocker.className = 'mejs__time-slider-blocker';
 
-    // Setup flash animation
     const flashBlockedOverlay = () => {
       scrubBlocker.classList.add('mejs__time-slider-blocker-error');
       setTimeout(() => {
@@ -571,14 +569,12 @@ class MediaView extends ComponentView {
       }, 150);
     };
 
-    // Add interaction handling for blocked area
     scrubBlocker.addEventListener('pointerdown', (e) => {
       e.preventDefault();
       e.stopImmediatePropagation();
       flashBlockedOverlay();
     });
 
-    // Add click handler to the entire slider for navigation to maxViewed
     sliderElement.addEventListener('click', (e) => {
       const rect = sliderElement.getBoundingClientRect();
       const clickX = e.clientX - rect.left;
@@ -599,7 +595,6 @@ class MediaView extends ComponentView {
       }
     });
 
-    // Append to slider
     sliderElement.style.position = 'relative';
     sliderElement.appendChild(scrubBlocker);
     sliderElement.setAttribute('aria-disabled', 'true');
@@ -624,7 +619,6 @@ class MediaView extends ComponentView {
         player.currentTime = getMaxViewed();
         setSuppress(false);
 
-        // Flash the blocker to indicate restriction
         scrubBlocker.classList.add('mejs__time-slider-blocker-error');
         setTimeout(() => {
           scrubBlocker.classList.remove('mejs__time-slider-blocker-error');
@@ -640,10 +634,8 @@ class MediaView extends ComponentView {
       if (forwardKeys.includes(e.code) && player.currentTime >= getMaxViewed()) {
         e.preventDefault();
 
-        // Navigate to maxViewed when trying to use forward keys
         player.currentTime = getMaxViewed();
 
-        // Flash the blocker
         scrubBlocker.classList.add('mejs__time-slider-blocker-error');
         setTimeout(() => {
           scrubBlocker.classList.remove('mejs__time-slider-blocker-error');
@@ -651,7 +643,6 @@ class MediaView extends ComponentView {
       }
     });
 
-    // Cleanup when media ends
     player.addEventListener('ended', () => {
       this.model.set('_isComplete', true);
       scrubBlocker.remove();
--- a/js/mediaView.js
+++ b/js/mediaView.js
@@ -536,10 +536,129 @@ class MediaView extends ComponentView {
    */
   preventForwardScrubbing() {
     if (!this.model.get('_preventForwardScrubbing') || this.model.get('_isComplete')) return;
-    $(this.mediaElement).on({
-      seeking: this.onMediaElementSeeking,
-      timeupdate: this.onMediaElementTimeUpdate
+
+    const player = this.mediaElement;
+    const $slider = this.$('.mejs__time-slider');
+    if (!$slider.length) return;
+
+    let maxViewed = this.model.get('_maxViewed') || 0;
+    let suppress = false;
+
+    // Create and setup the scrub blocker
+    const scrubBlocker = this.createScrubBlocker($slider[0]);
+
+    // Setup event handlers
+    this.setupScrubBlockerEvents(player, scrubBlocker, () => maxViewed, (newMax) => {
+      maxViewed = newMax;
+      this.model.set('_maxViewed', maxViewed);
+    }, () => suppress, (newSuppress) => {
+      suppress = newSuppress;
     });
+
+    // Initialize the blocker size
+    this.updateScrubBlocker(scrubBlocker, player, maxViewed);
+  }
+
+  createScrubBlocker(sliderElement) {
+    const scrubBlocker = document.createElement('div');
+    scrubBlocker.className = 'mejs__time-slider-blocker';
+
+    // Setup flash animation
+    const flashBlockedOverlay = () => {
+      scrubBlocker.classList.add('mejs__time-slider-blocker-error');
+      setTimeout(() => {
+        scrubBlocker.classList.remove('mejs__time-slider-blocker-error');
+      }, 150);
+    };
+
+    // Listen to the entire slider to detect blocked area clicks
+    sliderElement.addEventListener('click', (e) => {
+      const rect = sliderElement.getBoundingClientRect();
+      const clickX = e.clientX - rect.left;
+      const sliderWidth = rect.width;
+      const clickPercent = clickX / sliderWidth;
+
+      const player = this.mediaElement;
+      const duration = player.duration;
+      const clickTime = clickPercent * duration;
+      const maxViewed = this.model.get('_maxViewed') || 0;
+
+      // If clicking in the blocked area (beyond maxViewed)
+      if (clickTime > maxViewed + 0.25) {
+        e.preventDefault();
+        e.stopImmediatePropagation();
+
+        // Navigate to maxViewed and flash
+        player.currentTime = maxViewed;
+        flashBlockedOverlay();
+      }
+    });
+
+    // Append to slider
+    sliderElement.style.position = 'relative';
+    sliderElement.appendChild(scrubBlocker);
+    sliderElement.setAttribute('aria-disabled', 'true');
+
+    return scrubBlocker;
+  }
+
+  setupScrubBlockerEvents(player, scrubBlocker, getMaxViewed, setMaxViewed, getSuppress, setSuppress) {
+    // Update progress and blocker size
+    player.addEventListener('timeupdate', () => {
+      if (!getSuppress()) {
+        const newMaxViewed = Math.max(getMaxViewed(), player.currentTime);
+        setMaxViewed(newMaxViewed);
+        this.updateScrubBlocker(scrubBlocker, player, newMaxViewed);
+      }
+    });
+
+    // Prevent forward seeking and navigate to maxViewed
+    player.addEventListener('seeking', () => {
+      if (player.currentTime > getMaxViewed() + 0.25) {
+        setSuppress(true);
+        player.currentTime = getMaxViewed();
+        setSuppress(false);
+
+        // Flash the blocker to indicate restriction
+        scrubBlocker.classList.add('mejs__time-slider-blocker-error');
+        setTimeout(() => {
+          scrubBlocker.classList.remove('mejs__time-slider-blocker-error');
+        }, 150);
+
+        this._showBlockedScrubMessage?.();
+      }
+    });
+
+    // Prevent keyboard forward navigation
+    player.addEventListener('keydown', (e) => {
+      const forwardKeys = ['ArrowRight', 'End', 'PageDown'];
+      if (forwardKeys.includes(e.code) && player.currentTime >= getMaxViewed()) {
+        e.preventDefault();
+
+        // Navigate to maxViewed when trying to use forward keys
+        player.currentTime = getMaxViewed();
+
+        // Flash the blocker
+        scrubBlocker.classList.add('mejs__time-slider-blocker-error');
+        setTimeout(() => {
+          scrubBlocker.classList.remove('mejs__time-slider-blocker-error');
+        }, 150);
+      }
+    });
+
+    // Cleanup when media ends
+    player.addEventListener('ended', () => {
+      this.model.set('_isComplete', true);
+      scrubBlocker.remove();
+    });
+  }
+
+  updateScrubBlocker(scrubBlocker, player, maxViewed) {
+    const duration = player.duration;
+    if (!duration || duration === Infinity) return;
+
+    const percentViewed = 1 - maxViewed / duration;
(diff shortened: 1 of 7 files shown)


js/mediaView.js
```diff
@@ -536,10 +536,125 @@ class MediaView extends ComponentView {
...
+      if (!getSuppress()) {
+        const newMaxViewed = Math.max(getMaxViewed(), player.currentTime);
+        setMaxViewed(newMaxViewed);
+        this.updateScrubBlocker(scrubBlocker, player, newMaxViewed);
+      }
+    });
+
+    // Prevent forward seeking and navigate to maxViewed
+    player.addEventListener('seeking', () => {
+      if (player.currentTime > getMaxViewed() + 0.25) {
```
swashbuck, line 617: Could return early here

js/mediaView.js
```diff
@@ -536,10 +536,125 @@ class MediaView extends ComponentView {
...
+    sliderElement.appendChild(scrubBlocker);
+    sliderElement.setAttribute('aria-disabled', 'true');
+
+    return scrubBlocker;
+  }
+
+  setupScrubBlockerEvents(player, scrubBlocker, getMaxViewed, setMaxViewed, getSuppress, setSuppress) {
+  // Update progress and blocker size
+    player.addEventListener('timeupdate', () => {
+      if (!getSuppress()) {
```
swashbuck, line 608: Return early

js/mediaView.js
```diff
@@ -536,10 +536,125 @@ class MediaView extends ComponentView {
...
+      const sliderWidth = rect.width;
+      const clickPercent = clickX / sliderWidth;
+
+      const player = this.mediaElement;
+      const duration = player.duration;
+      const clickTime = clickPercent * duration;
+      const maxViewed = this.model.get('_maxViewed') || 0;
+
+      // If clicking ahead of maxViewed, navigate to maxViewed and flash
+      if (clickTime > maxViewed + 0.25) {
```
swashbuck, line 590: Return early

js/mediaView.js
```diff
@@ -536,10 +536,125 @@ class MediaView extends ComponentView {
...
+        }, 150);
+
+        this._showBlockedScrubMessage?.();
+      }
+    });
+
+    // Prevent keyboard forward navigation
+    player.addEventListener('keydown', (e) => {
+      const forwardKeys = ['ArrowRight', 'End', 'PageDown'];
+      if (forwardKeys.includes(e.code) && player.currentTime >= getMaxViewed()) {
```
swashbuck, line 634: Return early

js/mediaView.js
```diff
@@ -536,10 +536,125 @@ class MediaView extends ComponentView {
...
+  createScrubBlocker(sliderElement) {
+    const scrubBlocker = document.createElement('div');
+    scrubBlocker.className = 'mejs__time-slider-blocker';
+
+    const flashBlockedOverlay = () => {
+      scrubBlocker.classList.add('mejs__time-slider-blocker-error');
+      setTimeout(() => {
+        scrubBlocker.classList.remove('mejs__time-slider-blocker-error');
+      }, 150);
+    };
```
swashbuck, line 570: This code appears three times in this file. Suggest making it a function and calling that each time.

joe-replin: > This identical code appears three times in this file. Suggest making it a function and calling that each time.
Done. Good spot.

> Also, I'm wondering about the accessibility of the 150ms red flash. If someone clicks or key presses multiple times, the red can flash in quick succession. Maybe this should only flash a maximum of once every 1 second?
> 
> https://www.w3.org/WAI/WCAG22/Understanding/three-flashes-or-below-threshold.html#dfn-general-flash-and-red-flash-thresholds
I'm not so sure about this call. It's not a continuous animation and the user would need to rapid-fire click to create a violation of the rule, which is highly unlikely. Also, the goal of the rule is to avoid triggering seizures. The size of the square doesn't seem large enough to fall under WCAG 2.3.1? Happy to implement something but I'm no so sure it is needed. I could be swayed the other direction pretty easily though.

joe-replin:
Noting here that this PR has been extensively tested and is currently a part of a large production release with no issues.

less/mediaelementplayer.less
```diff
@@ -511,6 +511,23 @@ Reference: http://blog.rrwd.nl/2015/04/04/the-screen-reader-text-class-why-and-h
...
+  pointer-events: auto;
+  z-index: 9999;
+  cursor: not-allowed;
+  transition: background-color 0.15s ease;
+  
+  &.mejs__time-slider-blocker-error {
+    background-color: fade(@validation-error, 30%);
+  }
+}
+
```
swashbuck: Will you move these styles to _mep-overrides.less_? The _mediaelementplayer.less_ file is a direct copy of the library's [Less](https://github.com/mediaelement/mediaelement/blob/7.0.7/src/css/mediaelementplayer.css) file and should not be modified. Otherwise, the styles may be overridden during a library update.

joe-replin: Done!

cahirodoherty-learningpool (approved):
👀

js/mediaView.js
```diff
@@ -536,10 +536,115 @@ class MediaView extends ComponentView {
...
-    $(this.mediaElement).on({
-      seeking: this.onMediaElementSeeking,
-      timeupdate: this.onMediaElementTimeUpdate
+
+    const player = this.mediaElement;
+    const $slider = this.$('.mejs__time-slider');
+    if (!$slider.length) return;
+
+    let maxViewed = this.model.get('_maxViewed') ?? 0;
+    let suppress = false;
```
oliverfoster: These two variables and their associated functions seem completely mad, lots of unnecessarily complex code.

There are two callback functions each, passed into `setupScrubBlockerEvents`, `getMaxViewed`, `setMaxViewed`, `getSupress` and `setSupress`, both of which just toggle these local boolean variables.

Why not remove all four callback functions, the four arguments in `setupScrubBlockerEvents` and just set the boolean value on the MediaView instance as a `this.maxViewed` and `this.suppress` property instead?

oliverfoster (changes requested):

js/mediaView.js
```diff
@@ -536,10 +536,115 @@ class MediaView extends ComponentView {
...
+      setSuppress(false);
+
+      this.flashBlockedOverlay(scrubBlocker);
+
+      this._showBlockedScrubMessage?.();
+    });
+
+    // Prevent keyboard forward navigation
+    player.addEventListener('keydown', (e) => {
+      const forwardKeys = ['ArrowRight', 'End', 'PageDown'];
```
oliverfoster: This should be at the top of the file as `const FORWARD_KEYS = ['ArrowRight', 'End', 'PageDown'];` 

Currently, a keydown event occurs on the mediaelement or its descendants and then this array and its strings are created every time with identical values.

js/mediaView.js
```diff
@@ -536,10 +536,115 @@ class MediaView extends ComponentView {
    */
   preventForwardScrubbing() {
     if (!this.model.get('_preventForwardScrubbing') || this.model.get('_isComplete')) return;
-    $(this.mediaElement).on({
-      seeking: this.onMediaElementSeeking,
-      timeupdate: this.onMediaElementTimeUpdate
+
+    const player = this.mediaElement;
```
oliverfoster: The player variable does not need to be passed through to each function. As the functions are on the view, they can refer to `this.mediaElement`. This code is written as if the functions are not view instance functions.

oliverfoster (changes requested):

js/mediaView.js
joe-replin: Appreciate you calling me out on it. It was a convoluted approach. I will revise.

cahirodoherty-learningpool (approved):
👀

swashbuck (approved):
👍

js/mediaView.js
```diff
@@ -445,11 +447,39 @@ class MediaView extends ComponentView {
...
+      }
+      if (this._onScrubSeeking) {
+        this.mediaElement.removeEventListener('seeking', this._onScrubSeeking);
+      }
+      if (this._onScrubKeyDown) {
+        this.mediaElement.removeEventListener('keydown', this._onScrubKeyDown);
+      }
+      if (this._onScrubEnded) {
+        this.mediaElement.removeEventListener('ended', this._onScrubEnded);
+      }
```
oliverfoster: These checks are all a consequence of assigning the functions to the instance at runtime rather that assigning the functions to the class at compile time.

```suggestion
      this.mediaElement.removeEventListener('timeupdate', this._onScrubTimeUpdate);
      this.mediaElement.removeEventListener('seeking', this._onScrubSeeking);
      this.mediaElement.removeEventListener('keydown', this._onScrubKeyDown);
      this.mediaElement.removeEventListener('ended', this._onScrubEnded);
```

js/mediaView.js
```diff
@@ -535,11 +565,124 @@ class MediaView extends ComponentView {
...
+      this.mediaElement.currentTime = this._maxViewed;
+      this.flashBlockedOverlay(this._scrubBlocker);
+    };
+
+    this._onScrubEnded = () => {
+      this.model.set('_isComplete', true);
+      this._scrubBlocker?.remove();
+    };
+
+    this.mediaElement.addEventListener('timeupdate', this._onScrubTimeUpdate);
```
oliverfoster: All of these nested functions should be assigned to the media prototype at compile time rather than run-time.

```suggestion
    this._scrubBlocker = this.createScrubBlocker($slider[0]);

    this.setupScrubBlockerEvents();

    // Initialize the blocker size
    this.updateScrubBlocker();
  }
  
  _onBlockerPointerDown (e) {
    e.preventDefault();
    e.stopImmediatePropagation();
    this.flashBlockedOverlay(scrubBlocker);
  }

  _onSliderClick (e) {
    const rect = sliderElement.getBoundingClientRect();
    const clickX = e.clientX - rect.left;
    const sliderWidth = rect.width;
    const clickPercent = clickX / sliderWidth;
    const duration = this.mediaElement.duration;
    const clickTime = clickPercent * duration;
    const isClickingAhead = clickTime > this._maxViewed + 0.25;

    if (!isClickingAhead) return;

    e.preventDefault();
    e.stopImmediatePropagation();
    this.mediaElement.currentTime = this._maxViewed;
    this.flashBlockedOverlay(scrubBlocker);
  }

  createScrubBlocker(sliderElement) {
    const scrubBlocker = document.createElement('span');
    scrubBlocker.className = 'mejs__time-slider-blocker';

    scrubBlocker.addEventListener('pointerdown', this._onBlockerPointerDown);
    sliderElement.addEventListener('click', this._onSliderClick);

    sliderElement.style.position = 'relative';
    sliderElement.appendChild(scrubBlocker);
    sliderElement.setAttribute('aria-disabled', 'true');

    return scrubBlocker;
  }
  
  onScrubTimeUpdate () {
    if (this._suppressSeek) return;
    this._maxViewed = Math.max(this._maxViewed, this.mediaElement.currentTime);
    this.model.set('_maxViewed', this._maxViewed);
    this.updateScrubBlocker();
  }

  // Prevent forward seeking and navigate to maxViewed
  onScrubSeeking () {
    const isSeekingAhead = this.mediaElement.currentTime > this._maxViewed + 0.25;
    if (!isSeekingAhead) return;

    this._suppressSeek = true;
    this.mediaElement.currentTime = this._maxViewed;
    this._suppressSeek = false;

    this.flashBlockedOverlay(this._scrubBlocker);
    this._showBlockedScrubMessage?.();
  }

  // Prevent keyboard forward navigation
  _onScrubKeyDown (e) {
    const isForwardKey = FORWARD_SCRUBBING_KEYS.includes(e.code);
    const isAtMaxViewed = this.mediaElement.currentTime >= this._maxViewed;
    const shouldPrevent = isForwardKey && isAtMaxViewed;

    if (!shouldPrevent) return;

    e.preventDefault();
    this.mediaElement.currentTime = this._maxViewed;
    this.flashBlockedOverlay(this._scrubBlocker);
  }

  _onScrubEnded () {
    this.model.set('_isComplete', true);
    this._scrubBlocker?.remove();
  }

  setupScrubBlockerEvents() {
    // Update progress and blocker size
    this.mediaElement.addEventListener('timeupdate', this._onScrubTimeUpdate);
```

js/mediaView.js
```diff
@@ -535,11 +565,124 @@ class MediaView extends ComponentView {
...
+    this._onScrubSeeking = () => {
+      const isSeekingAhead = this.mediaElement.currentTime > this._maxViewed + 0.25;
+      if (!isSeekingAhead) return;
+
+      this._suppressSeek = true;
+      this.mediaElement.currentTime = this._maxViewed;
+      this._suppressSeek = false;
+
+      this.flashBlockedOverlay(this._scrubBlocker);
+      this._showBlockedScrubMessage?.();
```
oliverfoster: This function doesn't exist.
```
      this._showBlockedScrubMessage?.();
```

oliverfoster (changes requested):

js/mediaView.js
joe-replin: done

joe-replin:
Thank you for the good recommendations, @oliverfoster. It didn't even occur to me why moving all this to the class instead of the instance would be beneficial. Here's a list of changes I made for review:
## Fix
- Moved forward scrubbing prevention event handlers (`onScrubTimeUpdate`, `onScrubSeeking`, `onScrubKeyDown`, `onScrubEnded`) from runtime-assigned arrow function closures to class prototype methods bound in `preRender()` via `_.bindAll`
- Moved scrub blocker interaction handlers (`onBlockerPointerDown`, `onSliderClick`) from runtime closures to class prototype methods; slider element reference stored as `this._sliderElement` instead of captured via closure
- Moved captions change handler (`onCaptionsChange`) from an anonymous arrow function in `listenForCaptionsChange` to a class prototype method bound in `preRender()`
- Removed `if (this._onX)` existence guards from `remove()` — unconditional `removeEventListener` calls are now safe because all handlers are bound at initialisation
- Replaced direct `this.model.set('_isComplete', true)` in `onScrubEnded` with `setCompletionStatus()` to correctly use the Adapt completion API
- Renamed `flashBlockedOverlay` parameter from `e` to `element` to accurately reflect that it receives a DOM element, not an event
- Removed dead `onMediaElementSeeking` and `onMediaElementTimeUpdate` methods and their orphaned cleanup entries in `remove()`, along with the associated Android debounce workaround in `preRender()`
- Fixed `_suppressSeek` flag reset timing: flag is now cleared in a `{ once: true }` `seeked` event listener so it correctly suppresses `onScrubTimeUpdate` during the corrective seek

